PR #9, opened 2026-06-25
Nei config poc

sbancuz:
Apparently there is still some bug with the buttons for the NEI pages, will work on it when I have more time. This pr can still serve as a reference for nei configs


src/main/java/com/sbancuz/plannh/config/ConfigMain.java
TheYoingLad, line 1: Is this file still needed?

sbancuz: Not really, forgot to remove it

TheYoingLad (approved):

PR #14, opened 2026-07-05
Mui2 rework pt 2

TheYoingLad:
Group widget is almost done, some more robust children covering needs to be implemented/fixed on mui side.


src/main/java/com/sbancuz/plannh/gui/HeaderTextWidget.java
```diff
@@ -1,25 +1,33 @@
...
+    public HeaderTextWidget(FlowchartWidget<?, ?> parent, IntSupplier bgColor) {
         super(parent);
 
         height(20);
         setScale(1.5f);
         expanded();
-        background(new DynamicDrawable(() -> new Rectangle().color(isEditing ? bgColor : Color.argb(1f, 1f, 1f, 0))));
+        background(
+            new DynamicDrawable(
+                () -> new Rectangle().color(isEditing ? bgColor.getAsInt() : Color.argb(1f, 1f, 1f, 0))));
```
sbancuz, line 19: Should we define a default bg color for theming? I mean through the color class

src/main/java/com/sbancuz/plannh/config/ConfigOverrides.java
```diff
@@ -0,0 +1,14 @@
+package com.sbancuz.plannh.config;
+
+import com.gtnewhorizon.gtnhlib.config.Config;
+import com.sbancuz.plannh.PlanNH;
+
```
sbancuz: This file shouldn't exist, it was deleted from main

src/main/java/com/sbancuz/plannh/gui/HeaderTextWidget.java
TheYoingLad: There is already a default bg color for notes, this widget has configurable bg color on purpose.

For theming, the color from the Color class will be reused, but that will be addressed later

sbancuz (approved):
lgtm

PR #16, opened 2026-07-11
Add options for the canvas widget background

sbancuz:
Add a couple of options to the canvas background

- Gaussian Blur (with a setting for strength)
- Background color with color picker
- show/hide grid

All the setup for the shaders is just to add more types of blur or effects later (like kawase blur since it's more efficient)

### DIFF
--- a/src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
+++ b/src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
@@ -18,7 +18,9 @@ import com.cleanroommc.modularui.api.layout.IViewportStack;
 import com.cleanroommc.modularui.api.widget.IDraggable;
 import com.cleanroommc.modularui.api.widget.Interactable;
 import com.cleanroommc.modularui.drawable.BufferBuilder;
+import com.cleanroommc.modularui.drawable.DynamicDrawable;
 import com.cleanroommc.modularui.drawable.GuiDraw;
+import com.cleanroommc.modularui.drawable.Rectangle;
 import com.cleanroommc.modularui.drawable.Stencil;
 import com.cleanroommc.modularui.screen.ModularPanel;
 import com.cleanroommc.modularui.screen.viewport.ModularGuiContext;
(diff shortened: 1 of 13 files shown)


src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
```diff
@@ -293,14 +306,30 @@ private void addNodeWidget(final Node node) {
...
         final int aw = getArea().width;
         final int ah = getArea().height;
+        final int x = getArea().x;
+        final int y = getArea().y;
         if (aw <= 0 || ah <= 0) return;
 
-        // TODO add toggle for grid
-        drawGrid(aw, ah);
+        // Already in local coords
+        GuiDraw.drawRect(0, 0, aw, ah, backgroundColor.getHexValue(PlannhColors.BACKGROUND.getColor()));
```
TheYoingLad: Remove this and add `background(new DynamicDrawable(() -> new Rectangle().color(getBackgroundColor())))` to the ctor of CanvasWidget.

TheYoingLad (changes requested):
One comment on how the bg color was impleneted, other stuff looks good.

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
sbancuz: Is it because of resizing?

sbancuz: Should I create a custom IDrawable just for the background? Otherwise I would have to move the super.draw call and it might cause issues

TheYoingLad: > Is it because of resizing?

No, its because thats the mui2 way of doing backgrounds. You dont have to set the size nor do any draw calls.

This is how it will look with argb colors of (255, 0, 0, 100):

TheYoingLad: > Should I create a custom IDrawable just for the background? Otherwise I would have to move the super.draw call and it might cause issues

You dont need to do that

TheYoingLad (approved):

PR #17, opened 2026-07-13
Order's QOL Patches

OrderedSet86:
NOTE: this PR targets `balancer` branch as this is what I based the changes on; the `master` branch doesn't have functioning node rendering.

NOTE: This PR message is human made (I won't subject you to AI PR summaries), but the code in this PR is 100% AI generated. I did closely supervise it and had specific targets and test parameters for it to hit, but I can't fully attest to the actual Java code quality or existence of unwanted side effects. I am a professional Python dev, I know very little about Java. If it's any consolation, I did use Fable 5 for this, so perhaps it is good enough.

## Flowchart UI/UX

- Support R/U NEI keybinds when hovering ingredients in the flowchart
- When adding a recipe after using R/U from an ingredient in the flowchart, auto connect said recipe if it shares ingredients. This saves having to manually connect most of the time.
- When dragging an edge from a port manually, if the ingredient is a valid match to the currently hovered machine node, the ingredient snaps to the correct input without having to aim. Until mouse release, this is shown as a ghost selection (so the user can do something else if they want). In the image below, I am hovering the Gregtech logo.
image

- Make node ports and edges inherit color from the ingredient. Since NEI is the "source of truth" for what the end user thinks the ingredient looks like, render the ingredient in an off-screen framebuffer and use that for the sprite ground truth. To select the color from the sprite, the colors are cast to a 32x32x32 RGB cube and the modal color is selected. To avoid the sprite outline or transparency affecting the selected color, transparency is dropped and the edges are eroded by 4 pixels. Technical summary from Fable:
```
- **Derivation = render sampling:** the stack is rendered into a small
  offscreen framebuffer via NEI's own item drawing (the exact pipeline
  recipe screens use) and the modal color of the pixels is taken. This is
  immune to the icon API misrepresenting custom renderers (GT material
  items report white; GT blocks report unrelated casing textures; fluid
  display icons don't resolve to files). Details that mattered:
  - raw GL30 FBO with a *verified* binding — Minecraft's `Framebuffer`
    wrapper can silently no-op under Angelica, which made early versions
    sample the screen corner;
  - `GL_TEXTURE_2D` re-enabled before drawing (the sampler runs mid-GUI
    draw after untextured rect drawing);
  - the amount-text overlay suppressed (its white pixels won the vote);
  - sprite outlines eroded (pixels adjacent to transparency) and the
    histogram saturation-weighted, so pale gradient sprites (diamond)
    resolve to their hue instead of their white shine;
  - fallback chain for environments without GL30/rendering: per-pass
    sprite reads (GT material tint applied), resource-pack PNGs (Angelica
    frees the atlas' CPU-side frame data), fluid display stacks, and the
    old type colors as last resort.
- **Colors stay true to the sprite** (no luminance lift): readability on
  arbitrary world backgrounds comes from **contrast-adaptive outlines** —
  light outline under dark colors, dark under light — on both edges and
  pins, plus a wider dark underlay beneath edge lines and arrowheads.
```
image

- Set machine node titles to black for better contrast
image

- Fix flowchart zoom (not sure if you guys where having this issue, but the zoom on mine was slamming between 10% and 500%). Fable said this might be due to a MUI bug on dev installations:
```
- MUI2 passes the raw LWJGL wheel delta; real LWJGL2 reports
  ±120 per notch while lwjgl3ify (all modern GTNH packs) reports notch
  counts. Amounts ≥120 are now normalized by the LWJGL2 convention; pack
  behavior is byte-identical to before.
```
- Remove the port node label in favor of the usual NEI hover label. This may be a controversial change, so I can revert it if you like. My reasoning on this was that the old port label was having Z-fighting issues, sometimes it would render behind other nodes as it was part of the node rendering layer. From a UX perspective, the end user should never have to redesign their flowchart layout because they can't read a hover tooltip. This also frees up some space later to add edge quantities, which is in your TODO list.
image

- New nodes from NEI are now auto-placed rather than stacked on the (200, 200) origin. If the ingredient is an input, it goes on the left, if it's an output, it goes on the right. Auto-placed machine nodes are slightly offset to avoid edge collisions, and nodes are placed lower if there are existing nodes in the way

## Developer QOL

- Enforce single edge between port pairs
- Add -PgtnhRecipes flag to `./gradlew runClient` - this makes evaluating the balancer much easier
- Right now the derived colors are logged to `logs/fml-client-latest.log` for development purposes - this should be removed before 1.0 reease

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -58,4 +58,12 @@ dependencies {
 
     // ILP solver
     implementation('org.ojalgo:ojalgo:47.3.1')
+
+    // Opt-in GTNH pack recipes for the dev client: `./gradlew runClient -PgtnhRecipes`.
+    // Loads the GTNH coremod (dreamcraft) and its transitive mods so NEI shows the pack's
+    // recipe tweaks in dev instead of base GT5U recipes. Off by default to keep plain
+    // runClient light.
+    if (project.hasProperty('gtnhRecipes')) {
+        runtimeOnlyNonPublishable('com.github.GTNewHorizons:NewHorizonsCoreMod:2.9.13:dev')
+    }
 }
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -46,7 +46,7 @@ dependencies {
     compileOnly("net.glease:tc4recipelib:1.5.37:dev")
     compileOnly("com.github.GTNewHorizons:ForestryMC:4.11.27:dev")
     compileOnly("ganymedes01.etfuturum:Et-Futurum-Requiem:2.6.45-GTNH:dev")
-    devOnlyNonPublishable("com.github.GTNewHorizons:GT5-Unofficial:5.09.52.613:dev")
+    implementation("com.github.GTNewHorizons:GT5-Unofficial:5.09.52.613:dev")
     runtimeOnlyNonPublishable('com.github.GTNewHorizons:Angelica:2.1.38:dev') { transitive = false }
 
     // Speeds up mod identification and loading in dev
@@ -61,8 +61,7 @@ dependencies {
 
     // Opt-in GTNH pack recipes for the dev client: `./gradlew runClient -PgtnhRecipes`.
     // Loads the GTNH coremod (dreamcraft) and its transitive mods so NEI shows the pack's
-    // recipe tweaks in dev instead of base GT5U recipes. Off by default to keep plain
-    // runClient light.
+    // recipe tweaks in dev instead of base GT5U recipes. Useful for flowchart behavior checking
     if (project.hasProperty('gtnhRecipes')) {
         runtimeOnlyNonPublishable('com.github.GTNewHorizons:NewHorizonsCoreMod:2.9.13:dev')
     }
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -60,8 +60,7 @@ dependencies {
     implementation('org.ojalgo:ojalgo:47.3.1')
 
     // Opt-in GTNH pack recipes for the dev client: `./gradlew runClient -PgtnhRecipes`.
-    // Loads the GTNH coremod (dreamcraft) and its transitive mods so NEI shows the pack's
-    // recipe tweaks in dev instead of base GT5U recipes. Useful for flowchart behavior checking
+    // Loads the GTNH coremod (dreamcraft) so NEI shows the pack's recipe tweaks in dev.
     if (project.hasProperty('gtnhRecipes')) {
         runtimeOnlyNonPublishable('com.github.GTNewHorizons:NewHorizonsCoreMod:2.9.13:dev')
     }
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -46,7 +46,7 @@ dependencies {
     compileOnly("net.glease:tc4recipelib:1.5.37:dev")
     compileOnly("com.github.GTNewHorizons:ForestryMC:4.11.27:dev")
     compileOnly("ganymedes01.etfuturum:Et-Futurum-Requiem:2.6.45-GTNH:dev")
-    implementation("com.github.GTNewHorizons:GT5-Unofficial:5.09.52.613:dev")
+    devOnlyNonPublishable("com.github.GTNewHorizons:GT5-Unofficial:5.09.52.613:dev")
     runtimeOnlyNonPublishable('com.github.GTNewHorizons:Angelica:2.1.38:dev') { transitive = false }
 
     // Speeds up mod identification and loading in dev
--- a/src/main/java/com/sbancuz/plannh/api/PlanAPI.java
+++ b/src/main/java/com/sbancuz/plannh/api/PlanAPI.java
@@ -21,7 +21,6 @@ import net.minecraft.util.StatCollector;
 import com.sbancuz.plannh.data.flowchart.Graph;
 import com.sbancuz.plannh.data.flowchart.Serializer;
 import com.sbancuz.plannh.data.flowchart.SlotSet;
-import com.sbancuz.plannh.nei.NodeLookupContext;
 
 import codechicken.nei.NEIClientConfig;
 import codechicken.nei.NEIClientUtils;
@@ -33,14 +32,6 @@ public final class PlanAPI {
 
     @Nullable
     private static SlotSet slotSet = null;
-
-    private static final NodeLookupContext LOOKUP_CONTEXT = new NodeLookupContext();
-
-    /** The session's pending NEI lookup origin (see {@link NodeLookupContext}). */
-    @Nonnull
-    public static NodeLookupContext lookupContext() {
-        return LOOKUP_CONTEXT;
-    }
 
     @Nonnull
     public static SlotSet getSlotSet() {
--- a/src/main/java/com/sbancuz/plannh/api/PlanAPI.java
+++ b/src/main/java/com/sbancuz/plannh/api/PlanAPI.java
@@ -21,6 +21,7 @@ import net.minecraft.util.StatCollector;
 import com.sbancuz.plannh.data.flowchart.Graph;
 import com.sbancuz.plannh.data.flowchart.Serializer;
 import com.sbancuz.plannh.data.flowchart.SlotSet;
+import com.sbancuz.plannh.nei.NodeLookupContext;
 
 import codechicken.nei.NEIClientConfig;
 import codechicken.nei.NEIClientUtils;
@@ -32,6 +33,14 @@ public final class PlanAPI {
 
     @Nullable
     private static SlotSet slotSet = null;
+
+    private static final NodeLookupContext LOOKUP_CONTEXT = new NodeLookupContext();
+
+    /** The session's pending NEI lookup origin (see {@link NodeLookupContext}). */
+    @Nonnull
+    public static NodeLookupContext lookupContext() {
+        return LOOKUP_CONTEXT;
+    }
 
     @Nonnull
     public static SlotSet getSlotSet() {
--- a/src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
+++ b/src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
@@ -45,7 +45,6 @@ public final class RecipePropertyAPI {
             s -> 31 * s.getItem()
                 .hashCode() + s.getItemDamage())
         .displayStackProvider(stack -> stack)
-        .lookupMatcher(ItemStack::isItemEqual)
         .colorProvider(IngredientColors::itemColor)
         .pinInputColor(PlannhColors.PIN_INPUT.getColor())
         .pinOutputColor(PlannhColors.PIN_OUTPUT.getColor())
@@ -65,14 +64,9 @@ public final class RecipePropertyAPI {
         .hashCodeExtractor(
             fs -> fs.getFluid()
                 .hashCode())
-        // Guards sit inside the lambdas so GT classes only load if GT is present AND the
+        // Guard sits inside the lambda so GT classes only load if GT is present AND the
         // lambda actually runs (lambda bodies resolve their classes at call time, not here).
         .displayStackProvider(fs -> Compat.GREGTECH.isLoaded ? GTHooks.fluidDisplayStack(fs) : null)
-        .lookupMatcher((fs, lookup) -> {
-            if (!Compat.GREGTECH.isLoaded) return false;
-            final FluidStack lookupFluid = GTHooks.fluidFromDisplayStack(lookup);
-            return lookupFluid != null && lookupFluid.getFluid() == fs.getFluid();
-        })
         .colorProvider(IngredientColors::fluidColor)
         .pinInputColor(PlannhColors.PIN_FLUID_IN.getColor())
         .pinOutputColor(PlannhColors.PIN_FLUID_OUT.getColor())
--- a/src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
+++ b/src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
@@ -14,9 +14,13 @@ import net.minecraftforge.fluids.FluidRegistry;
 import net.minecraftforge.fluids.FluidStack;
 import net.minecraftforge.oredict.OreDictionary;
 
+import com.sbancuz.plannh.Compat;
 import com.sbancuz.plannh.data.PropertyProvider;
 import com.sbancuz.plannh.data.RecipeProperty;
 import com.sbancuz.plannh.data.RecipeResource;
+import com.sbancuz.plannh.data.provider.gregtech.GTHooks;
+import com.sbancuz.plannh.gui.IngredientColors;
+import com.sbancuz.plannh.gui.PlannhColors;
 
 public final class RecipePropertyAPI {
 
@@ -40,6 +44,12 @@ public final class RecipePropertyAPI {
         .hashCodeExtractor(
             s -> 31 * s.getItem()
                 .hashCode() + s.getItemDamage())
+        .displayStackProvider(stack -> stack)
+        .lookupMatcher(ItemStack::isItemEqual)
+        .colorProvider(IngredientColors::itemColor)
+        .pinInputColor(PlannhColors.PIN_INPUT.getColor())
+        .pinOutputColor(PlannhColors.PIN_OUTPUT.getColor())
+        .arrowColor(PlannhColors.ARROW_ITEM.getColor())
         .build();
 
     public static final RecipeResource<FluidStack> FLUID = RecipeResource
@@ -55,6 +65,18 @@ public final class RecipePropertyAPI {
         .hashCodeExtractor(
             fs -> fs.getFluid()
                 .hashCode())
+        // Guards sit inside the lambdas so GT classes only load if GT is present AND the
+        // lambda actually runs (lambda bodies resolve their classes at call time, not here).
+        .displayStackProvider(fs -> Compat.GREGTECH.isLoaded ? GTHooks.fluidDisplayStack(fs) : null)
+        .lookupMatcher((fs, lookup) -> {
+            if (!Compat.GREGTECH.isLoaded) return false;
+            final FluidStack lookupFluid = GTHooks.fluidFromDisplayStack(lookup);
+            return lookupFluid != null && lookupFluid.getFluid() == fs.getFluid();
+        })
+        .colorProvider(IngredientColors::fluidColor)
+        .pinInputColor(PlannhColors.PIN_FLUID_IN.getColor())
+        .pinOutputColor(PlannhColors.PIN_FLUID_OUT.getColor())
+        .arrowColor(PlannhColors.ARROW_FLUID.getColor())
         .build();
 
     private static boolean itemsMatch(final ItemStack a, final ItemStack b) {
--- a/src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
+++ b/src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
@@ -10,6 +10,7 @@ import javax.annotation.Nullable;
 
 import net.minecraft.init.Blocks;
 import net.minecraft.item.ItemStack;
+import net.minecraftforge.fluids.FluidContainerRegistry;
 import net.minecraftforge.fluids.FluidRegistry;
 import net.minecraftforge.fluids.FluidStack;
 import net.minecraftforge.oredict.OreDictionary;
@@ -66,7 +67,17 @@ public final class RecipePropertyAPI {
                 .hashCode())
         // Guard sits inside the lambda so GT classes only load if GT is present AND the
         // lambda actually runs (lambda bodies resolve their classes at call time, not here).
-        .displayStackProvider(fs -> Compat.GREGTECH.isLoaded ? GTHooks.fluidDisplayStack(fs) : null)
+        .displayStackProvider(fs -> {
+            if (Compat.GREGTECH.isLoaded) {
+                final ItemStack display = GTHooks.fluidDisplayStack(fs);
+                if (display != null) return display;
+            }
+            // Forge's registry fills only from a full container's worth, so normalize the
+            // amount; null when the fluid has no registered container.
+            return FluidContainerRegistry.fillFluidContainer(
+                new FluidStack(fs.getFluid(), FluidContainerRegistry.BUCKET_VOLUME),
+                FluidContainerRegistry.EMPTY_BUCKET.copy());
+        })
         .colorProvider(IngredientColors::fluidColor)
         .pinInputColor(PlannhColors.PIN_FLUID_IN.getColor())
         .pinOutputColor(PlannhColors.PIN_FLUID_OUT.getColor())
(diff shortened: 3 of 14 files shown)


src/main/java/com/sbancuz/plannh/gui/IngredientColors.java
```diff
@@ -0,0 +1,483 @@
...
+ * channel and the average color of the fullest bin wins. This tracks the dominant material color
+ * (e.g. copper orange) instead of being washed out by outlines and shading the way a plain
+ * average is.
+ */
+public final class IngredientColors {
+
+    private static final Map<String, Integer> CACHE = new HashMap<>();
+    private static final int ALPHA_OPAQUE_MIN = 128;
+
+    private static final int OUTLINE_DARK = 0xF00A0A0A;
```
sbancuz: Only use plannh colors

src/main/java/com/sbancuz/plannh/gui/IngredientColors.java
```diff
@@ -0,0 +1,483 @@
...
+                // pixels. This is immune to the icon API misrepresenting custom renderers
+                // (GT material items, GT blocks, fluid display items...).
+                final long[] rendered = sampleRenderedStack(stack);
+                if (rendered != null) return logColor(k, (int) rendered[0], "render-sample");
+
+                // weighted by how many opaque pixels each contributes.
+                final int passes = item.requiresMultipleRenderPasses()
+                    ? Math.max(1, item.getRenderPasses(stack.getItemDamage()))
+                    : 1;
+                final int gtTint = gtMaterialTint(stack);
```
sbancuz: Make sure the mod is compilable without GT, (so with compileOnly)

src/main/java/com/sbancuz/plannh/gui/IngredientColors.java
```diff
@@ -0,0 +1,483 @@
...
+        }
+    }
+
+    /**
+     * Chemistry-convention colors for fluids whose in-game color sources are unusable. Hydrogen
+     * defines a white fluid color, its texture region on the stitched atlas is empty, its fluid
+     * display item renders no pixels, and its GT material color is blue - while every player
+     * knows hydrogen as red-pink.
+     */
+    private static final Map<String, Integer> FLUID_CONVENTION_COLORS = Map.of("hydrogen", 0xE05A5A);
```
sbancuz: No overrides, if in the full pack we see differeneces then we'll deal with it later (try adding galacticraft/posttea and other relevant mods to check if the problem goes away)

src/main/java/com/sbancuz/plannh/gui/IngredientColors.java
```diff
@@ -0,0 +1,483 @@
...
+            final Integer convention = FLUID_CONVENTION_COLORS.get(fluid.getName());
+            if (convention != null) return logColor(key, convention, "convention");
+
+            // The fluid's own color if it defines one.
+            final int tint = fluid.getColor(fluidStack);
+            if ((tint & 0xFFFFFF) != 0xFFFFFF) return logColor(key, tint & 0xFFFFFF, "fluid-color");
+
+            // Otherwise color it the way NEI displays it: GT's fluid display item carries the
+            // exact icon + tint combination the player sees in recipe screens.
+            if (ModularUI.Mods.GT5U.isLoaded()) {
```
sbancuz: Nope, use the out mods isLoaded methods

src/main/java/com/sbancuz/plannh/gui/IngredientColors.java
```diff
@@ -0,0 +1,483 @@
...
+
+    /** True if a 4-neighbor is transparent; out-of-bounds neighbors count as opaque. */
+    private static boolean touchesTransparency(final int[] pixels, final int width, final int height, final int x,
+        final int y) {
+        return isTransparent(pixels, width, height, x - 1, y) || isTransparent(pixels, width, height, x + 1, y)
+            || isTransparent(pixels, width, height, x, y - 1)
+            || isTransparent(pixels, width, height, x, y + 1);
+    }
+
+    private static boolean isTransparent(final int[] pixels, final int width, final int height, final int x,
```
sbancuz, line 422: There is Color class to provide these methods

src/main/java/com/sbancuz/plannh/gui/RecipeNodeWidget.java
```diff
@@ -254,7 +272,7 @@ public void draw(final ModularGuiContext context, final WidgetThemeEntry<?> widg
                 (float) neiWidget.w / 2 - (float) titleW / 2,
                 TITLE_TEXT_Y,
                 1.0f,
-                PlannhColors.TEXT_WHITE.getColor(),
+                PlannhColors.TEXT_BLACK.getColor(),
```
sbancuz: The plan is to make it more dynamic since the header color is generated form the recipe handler name

src/main/java/com/sbancuz/plannh/nei/NodeLookupContext.java
```diff
@@ -0,0 +1,37 @@
...
+ * Remembers which node (and which of its ingredients) the user last ran an NEI recipe/usage
+ * lookup from, so a recipe added from that lookup can be wired back to the originating node
+ * automatically.
+ */
+public final class NodeLookupContext {
+
+    @Nullable
+    private static UUID nodeId;
+    @Nullable
+    private static ItemStack stack;
```
sbancuz: Does it handle fluids correctly? We only have itemstacks here

src/main/java/com/sbancuz/plannh/nei/NodeLookupContext.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private NodeLookupContext() {}
```
sbancuz: Why is everything static? Why are you using a singleton here?

src/main/java/com/sbancuz/plannh/nei/PlanOverlayHandler.java
```diff
@@ -63,6 +71,66 @@ private static void addRecipe(final GuiContainer firstGui, final IRecipeHandler
...
+     */
+    private static void autoConnectToLookupOrigin(final Graph graph, final Node added) {
+        final UUID originId = NodeLookupContext.nodeId();
+        final ItemStack lookup = NodeLookupContext.stack();
+        if (originId == null || lookup == null) return;
+        final Node origin = graph.nodes.get(originId);
+        if (origin == null || origin.id.equals(added.id)) return;
+
+        // R lookup: the new recipe produces what the origin consumes...
+        if (connectOnIngredient(graph, added, origin, lookup)) return;
```
sbancuz: This seems fishy without boolean flags specifying R/U

src/main/java/com/sbancuz/plannh/nei/PlanOverlayHandler.java
```diff
@@ -63,6 +71,66 @@ private static void addRecipe(final GuiContainer firstGui, final IRecipeHandler
...
+        }
+        return false;
+    }
+
+    private static boolean portMatchesLookup(final Port<?> port, final ItemStack lookup) {
+        final Object value = port.getValue();
+        if (value instanceof final ItemStack stack) {
+            return stack.isItemEqual(lookup);
+        }
+        if (value instanceof final FluidStack fluidValue && ModularUI.Mods.GT5U.isLoaded()) {
```
sbancuz: Same here

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
```diff
@@ -729,13 +740,37 @@ public void onMouseDrag(final int mouseButton, final long timeSinceClick) {
...
+    }
+
     @Override
     public boolean onMouseScroll(final UpOrDown direction, final int amount) {
         // Close context menu on any scroll ?
         // menuOpen = false;
 
         float delta = direction == UpOrDown.UP ? ZOOM_STEP : -ZOOM_STEP;
-        delta *= Math.max(1, Math.abs(amount));
+        // Real LWJGL2 (dev client, vanilla installs) reports the wheel in multiples of +-120 per
```
sbancuz: Comments like these can go away "Real LWJGL2"

sbancuz (changes requested):

src/main/java/com/sbancuz/plannh/api/PlanAPI.java
```diff
@@ -33,6 +34,14 @@ public final class PlanAPI {
     @Nullable
     private static SlotSet slotSet = null;
 
+    private static final NodeLookupContext LOOKUP_CONTEXT = new NodeLookupContext();
```
sbancuz: This is still a static singleton, if you want it, please explain a reason

src/main/java/com/sbancuz/plannh/gui/IngredientColors.java
```diff
@@ -462,22 +451,16 @@ private static int logColor(final String key, final int rgb, final String source
...
     private static int gtMaterialTint(final ItemStack stack) {
-        if (!ModularUI.Mods.GT5U.isLoaded()) return 0xFFFFFF;
-        if (stack.getItem() instanceof final MetaGeneratedItem gtItem) {
-            final short[] rgba = gtItem.getRGBa(stack);
-            if (rgba != null && rgba.length >= 3) {
-                return (rgba[0] & 0xFF) << 16 | (rgba[1] & 0xFF) << 8 | (rgba[2] & 0xFF);
-            }
-        }
-        return 0xFFFFFF;
+        return Compat.GREGTECH.isLoaded ? GTHooks.materialTint(stack) : NO_TINT;
```
sbancuz: Please no one line methods like this

src/main/java/com/sbancuz/plannh/nei/NodeLookupContext.java
```diff
@@ -9,29 +9,36 @@
...
-    private static ItemStack stack;
-
-    private NodeLookupContext() {}
-
-    public static void set(final UUID originNodeId, final ItemStack lookupStack) {
-        nodeId = originNodeId;
-        stack = lookupStack;
-    }
+    /** The node and ingredient an NEI lookup started from. */
+    public record Origin(UUID nodeId, ItemStack stack) {}
```
sbancuz: Why the nested record? This class could very well be the record

Also more in general, why do you need to save this? Why don't you call something when you press U/R with the itemstack?

dependencies.gradle
```diff
@@ -46,7 +46,7 @@ dependencies {
     compileOnly("net.glease:tc4recipelib:1.5.37:dev")
     compileOnly("com.github.GTNewHorizons:ForestryMC:4.11.27:dev")
     compileOnly("ganymedes01.etfuturum:Et-Futurum-Requiem:2.6.45-GTNH:dev")
-    implementation("com.github.GTNewHorizons:GT5-Unofficial:5.09.52.613:dev")
+    devOnlyNonPublishable("com.github.GTNewHorizons:GT5-Unofficial:5.09.52.613:dev")
```
sbancuz: ?? Revert this

src/main/java/com/sbancuz/plannh/nei/PlanOverlayHandler.java
sbancuz: This method is just port.canConnect, put an overload or something so we don't duplicate code

src/main/java/com/sbancuz/plannh/gui/FlowchartScreen.java
```diff
@@ -237,6 +241,33 @@ public void onClose() {
...
+    private void drawHoveredIngredientTooltip() {
+        if (!(getContext().getHovered() instanceof final RecipeViewerIngredientProvider provider)) return;
+        final ItemStack stack = provider.getStackForRecipeViewer();
+        if (stack == null) return;
+        final List<String> lines = codechicken.nei.guihook.GuiContainerManager
+            .itemDisplayNameMultiline(stack, null, true);
+        if (lines.isEmpty()) return;
+        GL11.glPushAttrib(GL11.GL_ENABLE_BIT | GL11.GL_DEPTH_BUFFER_BIT);
+        GL11.glDisable(GL11.GL_DEPTH_TEST);
+        codechicken.lib.gui.GuiDraw
```
sbancuz: No full paths

src/main/java/com/sbancuz/plannh/gui/RecipeNodeWidget.java
```diff
@@ -778,4 +789,103 @@ private void openNeiRecipe() {
...
+        }
+        if (port.getType() == RecipePropertyAPI.FLUID) {
+            final FluidStack fs = (FluidStack) port.getValue();
+            return fs != null && fs.amount > 0;
+        }
+        return false;
+    }
+
+    @Nullable
+    private static ItemStack stackForPort(final Port<?> port) {
```
sbancuz: No if conditions like these in the gui code. In recipe resource put a Function<> that can calculate the color from there, then you can just say port.getColor

src/main/java/com/sbancuz/plannh/gui/RecipeNodeWidget.java
```diff
@@ -778,4 +789,103 @@ private void openNeiRecipe() {
...
     }
+
+    /**
+     * Tells NEI what ingredient the mouse is over, which enables its native recipe/usage
+     * lookups (R/U and bookmark keys) on node contents: the embedded recipe's stacks, the
+     * throughput list rows, and the port pins.
+     */
+    @Override
+    @Nullable
+    public ItemStack getStackForRecipeViewer() {
```
sbancuz, line 785: This whole functionality feels overly engineered and won't scale when we will have more recipeResources, which is planned post 1.0

src/main/java/com/sbancuz/plannh/nei/PlanOverlayHandler.java
```diff
@@ -63,6 +75,76 @@ private static void addRecipe(final GuiContainer firstGui, final IRecipeHandler
...
+            return stack.isItemEqual(lookup);
+        }
+        if (value instanceof final FluidStack fluidValue && Compat.GREGTECH.isLoaded) {
+            final FluidStack lookupFluid = GTHooks.fluidFromDisplayStack(lookup);
+            return lookupFluid != null && lookupFluid.getFluid() == fluidValue.getFluid();
+        }
+        return false;
+    }
+
+    private static boolean hasIncomingEdge(final Graph graph, final UUID nodeId, final int inputIndex) {
```
sbancuz: Rewrite this method properly, you literally have the node id, why are you looping over every edge in the graph to get the edge?

sbancuz (changes requested):
Please respond to the actual comments if I ask for some clarification, don't just pass my comments to claude and push immediately the results without reading the code

src/main/java/com/sbancuz/plannh/gui/IngredientColors.java
OrderedSet86: I think LLMs in general are primarily trained on Javascript and Python; this kind of syntactic sugar is very common is JS so it's trying to apply it here. I assume the proper way of doing this is in functions so they can be overriden as needed by other mod devs?

sbancuz: Idk the proper way of doing this, but a one line function that gets called once to me is superfluous. Just inline the code

src/main/java/com/sbancuz/plannh/gui/IngredientColors.java
OrderedSet86: `isTransparent` is not in the MUI Color class, but I did switch to using methods from there instead of `>>>` in all the appropriate locations

src/main/java/com/sbancuz/plannh/gui/RecipeNodeWidget.java
OrderedSet86: I assume you mean R/U passing around an ItemStack, for that:

This is wired to the port index now instead of an ItemStack. When deciding whether the incoming machine node added from NEI should automatically get an edge, it calls:
```java
public boolean canConnect(final Port<?> other) {
    if (!type.equals(other.type)) return false;
    return ((RecipeResource<Object>) type).canConnect(value, other.value);
}
```
This allows for RecipeResource dependent behavior

src/main/java/com/sbancuz/plannh/nei/NodeLookupContext.java
OrderedSet86: Now this is:
```java
public record NodeLookupContext(UUID nodeId, boolean output, int portIndex) {}
```
The goal here is to auto-connect the new incoming machine node from NEI. The lookup ingredient is stored at U/R key press time, the actual edge addition can't be done until after the user selects the recipe, so some data needs to be stored in the short term

src/main/java/com/sbancuz/plannh/api/PlanAPI.java
OrderedSet86: This now lives inside `CanvasWidget`:
```java
@Nullable
private NodeLookupContext pendingLookup;

public void setPendingLookup(@Nullable final NodeLookupContext lookup) {
    pendingLookup = lookup;
}

/** Returns and clears the pending lookup: it wires at most one added recipe. */
@Nullable
public NodeLookupContext consumePendingLookup() {
    final NodeLookupContext result = pendingLookup;
    pendingLookup = null;
    return result;
}
```

OrderedSet86:
**New features / additions (will be added to the top PR summary):**
1. New nodes from NEI are now auto-placed rather than stacked on the (200, 200) origin. If the ingredient is an input, it goes on the left, if it's an output, it goes on the right. Auto-placed machine nodes are slightly offset to avoid edge collisions, and nodes are placed lower if there are existing nodes in the way
2. Enforce single edge between port pairs

src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
```diff
@@ -40,6 +44,12 @@ public final class RecipePropertyAPI {
         .hashCodeExtractor(
             s -> 31 * s.getItem()
                 .hashCode() + s.getItemDamage())
+        .displayStackProvider(stack -> stack)
+        .lookupMatcher(ItemStack::isItemEqual)
```
sbancuz: this is connectionChecker

src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
```diff
@@ -55,6 +65,18 @@ public final class RecipePropertyAPI {
         .hashCodeExtractor(
             fs -> fs.getFluid()
                 .hashCode())
+        // Guards sit inside the lambdas so GT classes only load if GT is present AND the
+        // lambda actually runs (lambda bodies resolve their classes at call time, not here).
+        .displayStackProvider(fs -> Compat.GREGTECH.isLoaded ? GTHooks.fluidDisplayStack(fs) : null)
```
sbancuz: I don't really like the null here, maybe use `FluidContainerRegistry.getFilledContainer()` as a fallback. I don't know if it would work, but in our context it's a really niche case, but at least won't throw an NPE

OrderedSet86:
There are a few items that don't have the proper edge/port color assigned (e.g. Block of Tiberium) but it works 90% of the time so I'm just going to leave it as is for now

src/main/java/com/sbancuz/plannh/gui/RecipeNodeWidget.java
```diff
@@ -459,22 +452,39 @@ private void drawThroughputInfo() {
...
-        y = drawPortList(x, y, node.inputs, nb, sec, ops, throughput, false);
-        drawPortList(x, y, node.outputs, nb, sec, ops, throughput, true);
-    }
-
-    private int drawPortList(final int x, int y, final List<Port<?>> ports, final NodeBalance nb, final float sec,
-        final int ops, final int throughput, final boolean output) {
-        for (int i = 0; i < ports.size(); i++) {
-            final Port port = ports.get(i);
-            final String label = portLabel(port, i, nb, sec, ops, throughput, output);
+        for (final ThroughputRow row : throughputRows()) {
```
sbancuz: Revert this change, drawPortList is more clear (at least to me)

src/main/java/com/sbancuz/plannh/nei/PlanOverlayHandler.java
```diff
@@ -81,70 +83,45 @@ private static void addRecipe(final GuiContainer firstGui, final IRecipeHandler
...
         final Node origin = graph.nodes.get(lookupOrigin.nodeId());
         final ItemStack lookup = lookupOrigin.stack();
         if (origin == null || lookup == null || origin.id.equals(added.id)) return;
 
-        final GuiScreen current = Minecraft.getMinecraft().currentScreen;
-        if (current instanceof GuiUsageRecipe) {
-            connectOnIngredient(graph, origin, added, lookup);
-        } else if (current instanceof GuiCraftingRecipe) {
-            connectOnIngredient(graph, added, origin, lookup);
+        if (connectOnIngredient(graph, added, origin, lookup)) {
```
sbancuz: I think the previous version was better, there could be recipes that have input/output with the same itemstack and I feel this would connect the wrong edge here

sbancuz (changes requested):
Overall way better than before, just a couple of nitpicks. Also check out if all the comments added are needed, or simply too long

PS: remeber that the canvas and flowchartScreen are mostly being rewritten so your first point could be already implemented

src/main/java/com/sbancuz/plannh/gui/RecipeNodeWidget.java
sbancuz: Damn, you read my mind

OrderedSet86: Agreed, I found this as you were reviewing. The complexity here was to support R/U lookups on throughput rows (under machine nodes), which is unnecessary and counterituitive UX

src/main/java/com/sbancuz/plannh/nei/PlanOverlayHandler.java
OrderedSet86: I think this was resolved during review as well

src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
OrderedSet86: also resolved during review lol
`lookupMatcher` got nuked in this commit: https://github.com/sbancuz/PlanNH/pull/17/changes/ac612a198be60cede7896c8ae40a057a766a44c9

src/main/java/com/sbancuz/plannh/nei/PlanOverlayHandler.java
sbancuz: Doesn't look like to me

src/main/java/com/sbancuz/plannh/api/RecipePropertyAPI.java
OrderedSet86: Fluids now default to bucket fill if GT is not loaded

src/main/java/com/sbancuz/plannh/nei/PlanOverlayHandler.java
OrderedSet86: oh I see what you mean. yeah there are a few weird recipes that output multiple itemstacks of the same thing, e.g. 3 stacks of 64 dusts. I'll have to see if I can find one

OrderedSet86: the issue here was actually an ingredient existing in both inputs and outputs, resolved by tracking the input side in `ingredientUnderMouse`

PR #18, opened 2026-07-18
Quick and Dirty fix so zooming doesn't break panning

HistyFisty:
Swapped the ordering of scale+pan so that zooming doesn't break things.


sbancuz (approved):
Thanks!

PR #19, opened 2026-07-19
Add auto-layout button

OrderedSet86:
## Functionality
- Adds auto-layout button, which uses SCC condensation height + ELK layout engine (new dependency)
- Add shift-click functionality to restore old chart if AL button was accidentally pressed
- Various improvements to ArrowRouter (never route under a machine node if there is an alternative option, clean up some pathological behavior, minimize bends)
- As a pre-emptive performance guard, graphs with >300 nodes use `BRANDES_KOEPF` instead of `NETWORK_SIMPLEX`. This I learned from trying to render nanocircuits in my flowv2 test repo. You lose some planarity, but it finishes in a reasonable amount of time
- For seamless user experience, warm up ELK in `ClientProxy.init` before first auto-layout

## Dev QOL
- Generalize port coordinate stuff into PortGeometry

## Downsides
- Adding ELK dependency adds 7MB to the publish jar. I cut down 1MB from it by dumping some unlikely to ever be used paths (docs, XML export), but there's only so much that can be done here. We will almost certainly use ELK in the long term (unless I figure out how to get OGDF working), so it's a good addition
- Changes more UI code, specifically `ArrowRouter`, making the inevitable MUI rework merge problem bigger (sorry)

## Licensing
| Library | Version | License |
|---|---|---|
| [Eclipse Layout Kernel (ELK)](https://github.com/eclipse-elk/elk) | 0.8.1 | [EPL-2.0](https://www.eclipse.org/legal/epl-2.0/) |
| [Eclipse Modeling Framework (EMF)](https://projects.eclipse.org/projects/modeling.emf.emf) - `common`, `ecore` | 2.12.0 | [EPL-1.0](https://www.eclipse.org/legal/epl-v10.html) |
| [Guava](https://github.com/google/guava) | 30.1-jre | [Apache-2.0](https://www.apache.org/licenses/LICENSE-2.0) |

## Jar Size Impacts
| Component | Uncompressed | Share | Outcome |
|---|---:|---:|---|
| guava (+annotations) | 7.77MB | 43% | pinned 33.6 → 30.1-jre (ELK declares the open range `[15.0,)`) |
| EMF (`ecore` 3.4 + `common` 0.75) | 4.14MB | 23% | kept; `ecore.xmi` (0.52MB) excluded — unused, graphs are built programmatically |
| ELK (`alg` 2.9, `core` 0.72, `graph` 0.22) | 3.86MB | 21% | kept — the actual layout engine |
| ELK's documentation images | 1.14MB | 6% | excluded (docs-site PNG/SVG illustrations) |
| `model/` ecore/xsd metamodels | 0.23MB | 1% | excluded (design-time EMF files) |
| `META-INF/versions` | 0.37MB | 2% | shrinks with the guava pin |
| PlanNH itself | ~0.6MB | 3% | |

## Demo Video

https://github.com/user-attachments/assets/c34446b4-23f9-465c-85cf-603db18fcd26

## Technical Explanation

(ai generated)

The sort is doing two jobs: making the layout **deterministic** (a function of the graph, not of the order recipes were clicked in NEI), and feeding ELK a **flow-aligned traversal order** so its cycle breaker cuts recycle loops in the right place.

### 1. ELK is only deterministic *given* a node order

<details>
<summary>Click to expand</summary>

We pin `RANDOM_SEED`, so ELK itself has no randomness left — but several of its phases consume nodes in **model order** (the order they were added to the `ElkNode`), and break ties by it. Without the sort, model order = NEI click order.

The clearest case is the cycle breaker we use (`CycleBreakingStrategy.DEPTH_FIRST`). From ELK 0.8.1's [`DepthFirstCycleBreaker`](https://github.com/eclipse-elk/elk/blob/master/plugins/org.eclipse.elk.alg.layered/src/org/eclipse/elk/alg/layered/p1cycles/DepthFirstCycleBreaker.java):

```java
for (LNode node : nodes) {                  // <- model order
    if (Iterables.isEmpty(node.getIncomingEdges())) {
        sources.add(node);
    }
}
for (LNode source : sources) {
    dfs(source);
}
// Start more DFS runs for all nodes that have not been visited yet.
// These must be part of a cycle since they are not source nodes
for (int i = 0; i < nodeCount; i++) {       // <- model order again
    if (!visited[i]) {
        dfs(nodes.get(i));
    }
}
```

Any edge the DFS finds pointing back into its current path gets reversed for layering. So for a **source-less loop** (very common here — self-sustaining recycle chains where every machine has an input), *which node happens to come first in model order is the DFS entry point, and that decides which edge gets reversed, which decides the entire left-to-right layering downstream.*

Concretely, take a loop `Electrolyzer → Reactor → Centrifuge → Electrolyzer` with a tap `Centrifuge → Chem Bath`. Every node has an incoming edge, so there are no sources and the DFS entry is pure model order:

```mermaid
flowchart LR
  subgraph A ["User clicked Reactor first → chart reads Reactor → Centrifuge → Electrolyzer"]
    R1[Reactor] --> C1[Centrifuge] --> E1[Electrolyzer]
    E1 -. reversed .-> R1
    C1 --> O1[Chem Bath]
  end
```

```mermaid
flowchart LR
  subgraph B ["User clicked Electrolyzer first → chart reads Electrolyzer → Reactor → Centrifuge"]
    E2[Electrolyzer] --> R2[Reactor] --> C2[Centrifuge]
    C2 -. reversed .-> E2
    C2 --> O2[Chem Bath]
  end
```

Same graph, two click histories, two different pictures. Worse: model order is just insertion order, so deleting and re-adding an unrelated node can reshuffle the whole chart on the next auto-layout. With the canonical sort, auto-layout is idempotent and identical for everyone who builds the same chart.

</details>

### 2. Why not something simpler, like sorting by UUID?

That would already make layout deterministic *for one saved chart* — but node UUIDs are random at creation, so the same logical chart rebuilt from scratch would sort differently and lay out differently. And it does nothing for the quality issue below. So the sort key has to come from the *structure* of the graph plus stable machine names; the id is only the last-resort tiebreak between identical machines at identical depth.

### 3. Why SCC-condensation height specifically

<details>
<summary>Click to expand</summary>

The order we actually want is "upstream machines first" — topological order. Then:

- ELK's DFS starts at true sources and walks *with* the flow, so the only edges it reverses are genuine recycle edges (instead of reversing a main-flow edge because the DFS happened to enter a loop from the side);
- on big graphs (>300 nodes) where we enable `CONSIDER_MODEL_ORDER`, ELK's model-order tie-breaking in layering/crossing minimization agrees with flow direction instead of click history;
- height 0 first = sources leftmost, matching `Direction.RIGHT`.

But a topological order doesn't exist for cyclic graphs, and recycle loops are the norm in recipe charts. The standard fix ([cp-algorithms has a good writeup](https://cp-algorithms.com/graph/strongly-connected-components.html)) is to contract each **strongly connected component** into one super-node. The resulting *condensation* is guaranteed to be a DAG, so it *can* be ordered — we take longest-path height from the sources:

```mermaid
flowchart LR
  subgraph S0 ["SCC 0 · height 0"]
    M[Miner]
  end
  subgraph S1 ["SCC 1 · height 1 — recycle loop"]
    Ma[Macerator] --> W[Ore Washer] --> Ce[Centrifuge] --> Ma
  end
  subgraph S2 ["SCC 2 · height 2"]
    EBF[Blast Furnace]
  end
  subgraph S3 ["SCC 3 · height 3"]
    As[Assembler]
  end
  M --> Ma
  Ce --> EBF
  EBF --> As
  style S0 fill:#d8f5d0,color:#000
  style S1 fill:#fbd2d2,color:#000
  style S2 fill:#fdf3c0,color:#000
  style S3 fill:#d3dcfb,color:#000
```

condenses to:

```mermaid
flowchart LR
  A0["{Miner}<br/>h = 0"] --> A1["{Macerator, Ore Washer, Centrifuge}<br/>h = 1"] --> A2["{Blast Furnace}<br/>h = 2"] --> A3["{Assembler}<br/>h = 3"]
  style A0 fill:#d8f5d0,color:#000
  style A1 fill:#fbd2d2,color:#000
  style A2 fill:#fdf3c0,color:#000
  style A3 fill:#d3dcfb,color:#000
```

Sort key is `(SCC height, name, id)`. All members of a loop share one height, so within a loop the order falls through to the name — which is exactly what makes the DFS entry point into each loop (case 1 above) deterministic too: it's always the alphabetically first machine in the loop, not whichever one was clicked first.

Tarjan is just the linear-time way to get the SCCs; it's written iteratively because recursion depth would be the chart's path length.

One more addition to avoid layout shifts. Now graph coloring is used to guarantee a unique solution based on graph metadata (neighbors etc). Also edges follow a canonical order as well.

</details>

### 4. Update: closing the last gap — the id tiebreak

<details>
<summary>Click to expand</summary>

Section 3 left one honest hole: the final id tiebreak. Ids are random at creation, so two machines tying on `(height, name)` still ordered randomly — the same logical chart rebuilt from scratch could swap them. The commit closes this with **wiring colors**: an iterated graph-coloring pass (color refinement) that identifies each node by its surroundings instead of its random id.

**Where the id used to leak through.** One Miner output feeds two Chemical Reactors (same machine, same size), one line continuing to a Blast Furnace, the other to a Distillery. Both reactors tie on `(height, name)`, and because both edges leave the *same output port*, both vertical stackings have equal edge crossings — a genuine tie that only model order breaks. So whichever reactor happened to draw the smaller id decided the picture; in a headless harness that rebuilds the chart with fresh random ids and shuffled insertion order, 17 of 39 rebuilds flipped it:

```mermaid
flowchart LR
  subgraph A ["Rebuild 1 — EBF branch happens to sort first"]
    M1[Miner] --> RA1[Chemical Reactor] --> E1[Blast Furnace]
    M1 --> RB1[Chemical Reactor] --> S1[Distillery]
  end
```

```mermaid
flowchart LR
  subgraph B ["Rebuild 2 — Distillery branch happens to sort first"]
    M2[Miner] --> RB2[Chemical Reactor] --> S2[Distillery]
    M2 --> RA2[Chemical Reactor] --> E2[Blast Furnace]
  end
```

**Links leak the same way.** Holding the nodes, the ids and the node order all fixed and shuffling *only* the order links were drawn in still flips the branches on 19 of 39 rebuilds — node positions, not just edge routes: with equal crossings and fixed ports, ELK keeps whichever order it encounters first, and encounter order follows the edge list. So links are also sorted to follow the node canonicalization — `(source position, output port, target position, input port)` — which keeps edge routing stable for free.

**The coloring.** To pin the node order down, every node gets a canonical color. Take a harder version of the chart above, where the twin branches stay identical for one more hop before diverging.

Round 0 — color nodes by what they *are*: `(size, port counts, height, name)`. All four Chemical Reactors are indistinguishable and share a color:

```mermaid
flowchart LR
  M0[Miner] --> A0[Chemical Reactor] --> CA0[Chemical Reactor] --> E0[Blast Furnace]
  M0 --> B0[Chemical Reactor] --> CB0[Chemical Reactor] --> S0[Distillery]
  style M0 fill:#d8f5d0,color:#000
  style A0 fill:#fbd2d2,color:#000
  style B0 fill:#fbd2d2,color:#000
  style CA0 fill:#fbd2d2,color:#000
  style CB0 fill:#fbd2d2,color:#000
  style E0 fill:#d3dcfb,color:#000
  style S0 fill:#fdf3c0,color:#000
```

Round 1 — recolor every node by its own color plus the sorted list of `(own port, direction, peer color)`. The two middle reactors split, because one's output peer is a Blast Furnace and the other's is a Distillery. The two left reactors still tie: both see a green Miner upstream and a red reactor downstream:

```mermaid
flowchart LR
  M3[Miner] --> A3[Chemical Reactor] --> CA3[Chemical Reactor] --> E3[Blast Furnace]
  M3 --> B3[Chemical Reactor] --> CB3[Chemical Reactor] --> S3[Distillery]
  style M3 fill:#d8f5d0,color:#000
  style A3 fill:#fbd2d2,color:#000
  style B3 fill:#fbd2d2,color:#000
  style CA3 fill:#e8d5f5,color:#000
  style CB3 fill:#fde3c8,color:#000
  style E3 fill:#d3dcfb,color:#000
  style S3 fill:#fdf3c0,color:#000
```

Round 2 — recolor again, now with the round-1 colors. The left reactors split too, because their downstream peers no longer match:

```mermaid
flowchart LR
  M4[Miner] --> A4[Chemical Reactor] --> CA4[Chemical Reactor] --> E4[Blast Furnace]
  M4 --> B4[Chemical Reactor] --> CB4[Chemical Reactor] --> S4[Distillery]
  style M4 fill:#d8f5d0,color:#000
  style A4 fill:#cdeef0,color:#000
  style B4 fill:#f5c6dc,color:#000
  style CA4 fill:#e8d5f5,color:#000
  style CB4 fill:#fde3c8,color:#000
  style E4 fill:#d3dcfb,color:#000
  style S4 fill:#fdf3c0,color:#000
```

Every node now has its own color, so one more round changes nothing and the refinement stops. In general, a difference k hops away separates twins after k rounds — which is why the pass iterates at all: a single round would have stopped after the middle diagram, left the two red reactors tied on the random id, and the harness shows exactly that chart flipping on 19 of 39 rebuilds.

Two details that make the colors canonical: within a round, each new color is the *rank of the node's signature in the sorted list of distinct signatures*, so colors depend only on the graph, never on iteration order; and the signature includes port indices, so "feeds input 0" and "feeds input 1" of the same peer are different wirings.

The sort key becomes `(height, name, color, id)`. After this change, the id correctly orders all but true structural twins — and with iteration, "twin" now means *no chain of links anywhere in the chart distinguishes the two nodes*, so swapping them is invisible on screen.

**Cost.** Each round builds signatures in O(V + E) and ranks them with one sort; every round before the fixpoint must split at least one color class, so rounds ≤ V, bounding the worst case by O(V·(V+E)·log V). In practice rounds = distance to the nearest distinguishing feature, 2–4 on real charts. As a deliberately adversarial measurement: a 500-node chain of identically named machines (forces exactly one hop of progress per round, ~500 rounds) runs the whole `layout()` call, ELK included, in ~310 ms.

**Results.** Three harness scenarios — same-port fan-out, disconnected same-name components, and the two-hop twins above — each rebuilt 40 times with fresh random ids and shuffled node/link insertion: all coordinate-identical. The existing reference charts lay out unchanged.

</details>

### DIFF
--- a/addon.gradle
+++ b/addon.gradle
@@ -1,0 +1,16 @@
+// Applied by gtnhgradle's AddonScriptModule alongside the standard convention scripts.
+
+// The shaded ELK distribution ships its documentation-site illustrations (images/**) and
+// EMF design-time metamodels (model/**); both are dead weight at runtime.
+tasks.matching { it.name == 'shadowJar' }.configureEach {
+    exclude 'images/**'
+    exclude 'model/**'
+}
+
+// ELK declares guava as the open range [15.0,), which re-resolves to the newest guava on
+// every fresh resolution - unreproducible builds and a needlessly fat shaded jar. Pin the
+// ELK-0.8-era release. Scoped to the shadow configurations so the dev/runtime classpath
+// (which must keep Minecraft's guava) is untouched.
+configurations.matching { it.name.toLowerCase(Locale.ROOT).contains('shadow') }.configureEach {
+    resolutionStrategy.force('com.google.guava:guava:30.1-jre')
+}
--- a/addon.gradle
+++ b/addon.gradle
@@ -1,10 +1,20 @@
 // Applied by gtnhgradle's AddonScriptModule alongside the standard convention scripts.
 
-// The shaded ELK distribution ships its documentation-site illustrations (images/**) and
-// EMF design-time metamodels (model/**); both are dead weight at runtime.
+// The shaded ELK distribution ships its documentation-site illustrations (images/**), EMF
+// design-time metamodels (model/** and schema/**), Eclipse plugin descriptors, and docs;
+// all dead weight at runtime, and the descriptors would sit at the jar's top level.
 tasks.matching { it.name == 'shadowJar' }.configureEach {
     exclude 'images/**'
     exclude 'model/**'
+    exclude 'schema/**'
+    exclude 'docs/**'
+    exclude 'about.*'
+    exclude 'plugin.xml'
+    exclude 'plugin.properties'
+    exclude 'modeling32.png'
+    exclude 'META-INF/maven/**'
+    exclude 'META-INF/LICENSE*'
+    exclude 'META-INF/NOTICE*'
 }
 
 // ELK declares guava as the open range [15.0,), which re-resolves to the newest guava on
--- a/addon.gradle
+++ b/addon.gradle
@@ -4,6 +4,11 @@
 // design-time metamodels (model/** and schema/**), Eclipse plugin descriptors, and docs;
 // all dead weight at runtime, and the descriptors would sit at the jar's top level.
 tasks.matching { it.name == 'shadowJar' }.configureEach {
+    // Three ELK jars each declare an ILayoutMetaDataProvider service. Without this, shadow keeps
+    // one of the three and leaves its filename and contents on the original package names, so a
+    // ServiceLoader in the relocated ELK finds nothing - in the shipped jar only, never in dev.
+    duplicatesStrategy = DuplicatesStrategy.INCLUDE
+    mergeServiceFiles()
     exclude 'images/**'
     exclude 'model/**'
     exclude 'schema/**'
--- a/src/main/java/com/sbancuz/plannh/ClientProxy.java
+++ b/src/main/java/com/sbancuz/plannh/ClientProxy.java
@@ -13,6 +13,7 @@ import com.sbancuz.plannh.client.ChatHandler;
 import com.sbancuz.plannh.client.ImportCommand;
 import com.sbancuz.plannh.client.WorldHandler;
 import com.sbancuz.plannh.gui.FlowchartScreen;
+import com.sbancuz.plannh.layout.AutoLayout;
 
 import cpw.mods.fml.client.registry.ClientRegistry;
 import cpw.mods.fml.common.FMLCommonHandler;
@@ -54,6 +55,13 @@ public class ClientProxy extends CommonProxy {
         FMLCommonHandler.instance()
             .bus()
             .register(this);
+
+        // ELK's first layout pays for its class loading and metadata registration; warm it up
+        // off-thread so the first Auto-Layout click doesn't freeze the game.
+        final Thread elkWarmup = new Thread(AutoLayout::warmup, "PlanNH ELK warmup");
+        elkWarmup.setDaemon(true);
+        elkWarmup.setPriority(Thread.MIN_PRIORITY);
+        elkWarmup.start();
     }
 
     @SubscribeEvent
--- a/src/main/java/com/sbancuz/plannh/ClientProxy.java
+++ b/src/main/java/com/sbancuz/plannh/ClientProxy.java
@@ -56,12 +56,9 @@ public class ClientProxy extends CommonProxy {
             .bus()
             .register(this);
 
-        // ELK's first layout pays for its class loading and metadata registration; warm it up
-        // off-thread so the first Auto-Layout click doesn't freeze the game.
-        final Thread elkWarmup = new Thread(AutoLayout::warmup, "PlanNH ELK warmup");
-        elkWarmup.setDaemon(true);
-        elkWarmup.setPriority(Thread.MIN_PRIORITY);
-        elkWarmup.start();
+        // ELK's first layout pays for its class loading and metadata registration; spend it
+        // during init so the first Auto-Layout click doesn't freeze the game.
+        AutoLayout.warmup();
     }
 
     @SubscribeEvent
--- a/src/main/java/com/sbancuz/plannh/ClientProxy.java
+++ b/src/main/java/com/sbancuz/plannh/ClientProxy.java
@@ -56,15 +56,7 @@ public class ClientProxy extends CommonProxy {
             .bus()
             .register(this);
 
-        // ELK's first layout pays for its class loading and metadata registration; spend it
-        // during init so the first Auto-Layout click doesn't freeze the game. AutoLayout's static
-        // block runs on this call, so a mis-shaded ELK fails here as a LinkageError - which must
-        // cost the mod its layout button, not its startup.
-        try {
-            AutoLayout.warmup();
-        } catch (final RuntimeException | LinkageError e) {
-            PlanNH.LOG.error("ELK warm-up failed; Auto-Layout will be unavailable", e);
-        }
+        Minecraft.getMinecraft().func_152344_a(AutoLayout::warmup);
     }
 
     @SubscribeEvent
--- a/src/main/java/com/sbancuz/plannh/ClientProxy.java
+++ b/src/main/java/com/sbancuz/plannh/ClientProxy.java
@@ -57,8 +57,14 @@ public class ClientProxy extends CommonProxy {
             .register(this);
 
         // ELK's first layout pays for its class loading and metadata registration; spend it
-        // during init so the first Auto-Layout click doesn't freeze the game.
-        AutoLayout.warmup();
+        // during init so the first Auto-Layout click doesn't freeze the game. AutoLayout's static
+        // block runs on this call, so a mis-shaded ELK fails here as a LinkageError - which must
+        // cost the mod its layout button, not its startup.
+        try {
+            AutoLayout.warmup();
+        } catch (final RuntimeException | LinkageError e) {
+            PlanNH.LOG.error("ELK warm-up failed; Auto-Layout will be unavailable", e);
+        }
     }
 
     @SubscribeEvent
--- a/src/main/java/com/sbancuz/plannh/ClientProxy.java
+++ b/src/main/java/com/sbancuz/plannh/ClientProxy.java
@@ -56,7 +56,8 @@ public class ClientProxy extends CommonProxy {
             .bus()
             .register(this);
 
-        Minecraft.getMinecraft().func_152344_a(AutoLayout::warmup);
+        Minecraft.getMinecraft()
+            .func_152344_a(AutoLayout::warmup);
     }
 
     @SubscribeEvent
--- a/src/main/java/com/sbancuz/plannh/gui/ArrowRouter.java
+++ b/src/main/java/com/sbancuz/plannh/gui/ArrowRouter.java
@@ -18,20 +18,41 @@ import java.util.UUID;
  * produced paths stable while the user pans (a pure translation) and lets the caller cache them.
  * <p>
  * The router lays a uniform grid over the bounding region and runs A* where the search state is
- * {@code (cell, incoming-direction)} so that turns can be penalized. Node rectangles inflated by a
- * margin are hard obstacles; cells already occupied by a routed arrow carry a soft penalty so that
- * later arrows take a parallel track instead of overlapping.
+ * {@code (cell, incoming-direction)} so that turns can be penalized. Node rectangles are hard
+ * obstacles; their surrounding margin ring and cells already occupied by a routed arrow carry soft
+ * penalties, so arrows keep clearance where there is room but can squeeze through tight gaps.
  */
 public final class ArrowRouter {
 
     /** Per-cell movement cost. */
     private static final int STEP = 1;
-    /** Extra cost for changing direction. Keeps the number of bends low. */
-    private static final int TURN = 12;
-    /** Extra cost per cell that already carries another arrow. Makes arrows separate. */
-    private static final int ARROW = 4;
+    /**
+     * Extra cost for changing direction. High relative to STEP so paths never micro-dodge
+     * around mild congestion - a single sideways jog costs two turns, which no small
+     * occupancy saving can justify.
+     */
+    private static final int TURN = 24;
+    /**
+     * Extra cost per cell that already carries another arrow. Kept low relative to STEP so
+     * that sharing even a long congested corridor stays cheaper than detouring around the
+     * outside of the chart.
+     */
+    private static final int ARROW = 2;
     /** Smaller penalty around an occupied cell, so arrows keep at least one cell of clearance. */
     private static final int NEAR = 1;
+    /**
+     * Penalty per cell inside a node's margin ring. The ring is deliberately soft, not blocked:
+     * nodes placed with a gap smaller than two margins would otherwise seal the gap shut and
+     * force absurd whole-chart detours. Arrows avoid hugging nodes when there is room, but can
+     * squeeze through tight gaps when there is not.
+     */
+    private static final int MARGIN_COST = 8;
+    /**
+     * Penalty for routing through another edge's port anchor (the straight approach run just
+     * right of an output pin / just left of an input pin). Anchors stay visually clear so every
+     * arrow's first and last segment reads as belonging to its port.
+     */
+    private static final int ANCHOR_COST = 30;
 
     /** Hard cap on grid cells; the cell size is grown if a region would exceed it. */
     private static final int MAX_CELLS = 200_000;
@@ -103,9 +124,17 @@ public final class ArrowRouter {
 
         final Grid grid = new Grid(minX, minY, cols, rows, cell, stub);
         grid.blockObstacles(obstacles, margin);
+        grid.reserveAnchors(requests);
 
-        for (final Request q : requests) {
-            final List<int[]> cellPath = grid.search(q);
+        for (int i = 0; i < requests.size(); i++) {
+            final Request q = requests.get(i);
+            // Ports closer than two anchor stubs cannot satisfy the leave-right/arrive-right
+            // state machine without looping around themselves; draw the canonical Z directly.
+            if (q.dx - q.sx < 2 * stub && Math.abs(q.dy - q.sy) < 10 * baseCell) {
+                result.put(q.key, fallback(q, stub));
+                continue;
+            }
+            final List<int[]> cellPath = grid.search(q, i);
             final List<int[]> path;
             if (cellPath == null) {
                 path = fallback(q, stub);
@@ -134,6 +163,7 @@ public final class ArrowRouter {
         final int originX, originY, cols, rows, cell, stub;
         final boolean[] blocked;
         final int[] occupancy;
+        final int[] anchorOwner;
         final int[] gScore;
         final int[] cameFrom;
         final int[] scoreRun;
@@ -148,6 +178,8 @@ public final class ArrowRouter {
             this.stub = stub;
             this.blocked = new boolean[cols * rows];
             this.occupancy = new int[cols * rows];
+            this.anchorOwner = new int[cols * rows];
+            Arrays.fill(anchorOwner, -1);
             final int states = cols * rows * 4;
             this.gScore = new int[states];
             this.cameFrom = new int[states];
@@ -174,11 +206,33 @@ public final class ArrowRouter {
             for (final Rect r : obstacles) {
                 final int x0 = gx(r.x - margin), x1 = gx(r.x + r.w + margin);
                 final int y0 = gy(r.y - margin), y1 = gy(r.y + r.h + margin);
+                final int bx0 = gx(r.x), bx1 = gx(r.x + r.w);
+                final int by0 = gy(r.y), by1 = gy(r.y + r.h);
                 for (int y = y0; y <= y1; y++) {
                     for (int x = x0; x <= x1; x++) {
-                        blocked[y * cols + x] = true;
+                        if (x >= bx0 && x <= bx1 && y >= by0 && y <= by1) {
+                            blocked[y * cols + x] = true;
+                        } else {
+                            occupancy[y * cols + x] += MARGIN_COST;
+                        }
                     }
                 }
+            }
+        }
+
+        /** Marks every request's port-approach runs so other arrows keep out of them. */
+        void reserveAnchors(final List<Request> requests) {
+            for (int i = 0; i < requests.size(); i++) {
+                final Request q = requests.get(i);
+                markAnchor(gx(q.sx), gx(q.sx + stub), gy(q.sy), i);
+                markAnchor(gx(q.dx - stub), gx(q.dx), gy(q.dy), i);
+            }
+        }
+
+        private void markAnchor(final int x0, final int x1, final int y, final int owner) {
+            for (int x = x0; x <= x1; x++) {
+                final int idx = y * cols + x;
+                if (anchorOwner[idx] == -1) anchorOwner[idx] = owner;
             }
         }
 
@@ -196,7 +250,7 @@ public final class ArrowRouter {
         }
 
         /** A* over (cell, direction); returns the list of {@code {gx, gy}} cells or null. */
-        List<int[]> search(final Request q) {
+        List<int[]> search(final Request q, final int requestIndex) {
             final int sgx = gx(q.sx + stub), sgy = gy(q.sy);
             final int ggx = gx(q.dx - stub), ggy = gy(q.dy);
             final int run = nextRun();
@@ -229,10 +283,11 @@ public final class ArrowRouter {
                     final int ny = cy + DY[nd];
                     if (nx < 0 || nx >= cols || ny < 0 || ny >= rows) continue;
                     final int nIdx = ny * cols + nx;
-                    final boolean isGoal = nx == ggx && ny == ggy;
-                    if (blocked[nIdx] && !isGoal) continue;
+                    if (blocked[nIdx]) continue;
 
-                    final int cost = STEP + (nd != dir ? TURN : 0) + occupancy[nIdx];
+                    final int foreignAnchor = anchorOwner[nIdx] != -1 && anchorOwner[nIdx] != requestIndex ? ANCHOR_COST
+                        : 0;
+                    final int cost = STEP + (nd != dir ? TURN : 0) + occupancy[nIdx] + foreignAnchor;
                     final int ng = g + cost;
                     final int nState = nIdx * 4 + nd;
                     if (ng < score(nState, run)) {
--- a/src/main/java/com/sbancuz/plannh/gui/ArrowRouter.java
+++ b/src/main/java/com/sbancuz/plannh/gui/ArrowRouter.java
@@ -2,6 +2,7 @@ package com.sbancuz.plannh.gui;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
@@ -26,31 +27,23 @@ public final class ArrowRouter {
 
     /** Per-cell movement cost. */
     private static final int STEP = 1;
-    /**
-     * Extra cost for changing direction. High relative to STEP so paths never micro-dodge
-     * around mild congestion - a single sideways jog costs two turns, which no small
-     * occupancy saving can justify.
-     */
+    /** Extra cost per direction change; high relative to STEP so paths don't micro-dodge. */
     private static final int TURN = 24;
     /**
-     * Extra cost per cell that already carries another arrow. Kept low relative to STEP so
-     * that sharing even a long congested corridor stays cheaper than detouring around the
-     * outside of the chart.
+     * Extra cost per cell already carrying another arrow; low, so sharing a long corridor
+     * stays cheaper than detouring around the chart.
      */
     private static final int ARROW = 2;
     /** Smaller penalty around an occupied cell, so arrows keep at least one cell of clearance. */
     private static final int NEAR = 1;
     /**
-     * Penalty per cell inside a node's margin ring. The ring is deliberately soft, not blocked:
-     * nodes placed with a gap smaller than two margins would otherwise seal the gap shut and
-     * force absurd whole-chart detours. Arrows avoid hugging nodes when there is room, but can
-     * squeeze through tight gaps when there is not.
+     * Penalty per cell in a node's margin ring. Soft rather than blocked: hard margins seal
+     * any gap narrower than two margins and force whole-chart detours.
      */
     private static final int MARGIN_COST = 8;
     /**
-     * Penalty for routing through another edge's port anchor (the straight approach run just
-     * right of an output pin / just left of an input pin). Anchors stay visually clear so every
-     * arrow's first and last segment reads as belonging to its port.
+     * Penalty for crossing another edge's port anchor (the straight approach run beside a
+     * pin), keeping each pin's first and last segment visually its own.
      */
     private static final int ANCHOR_COST = 30;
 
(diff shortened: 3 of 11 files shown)


OrderedSet86:
Sorry about the two force pushes, the change was removing the MUI logspam fix as it is now covered by https://github.com/sbancuz/PlanNH/pull/20. Should be correct now, fortunately I had a backup

src/main/java/com/sbancuz/plannh/gui/ArrowRouter.java
```diff
@@ -18,20 +19,33 @@
...
  * <p>
  * The router lays a uniform grid over the bounding region and runs A* where the search state is
- * {@code (cell, incoming-direction)} so that turns can be penalized. Node rectangles inflated by a
- * margin are hard obstacles; cells already occupied by a routed arrow carry a soft penalty so that
- * later arrows take a parallel track instead of overlapping.
+ * {@code (cell, incoming-direction)} so that turns can be penalized. Node rectangles are hard
+ * obstacles; their surrounding margin ring and cells already occupied by a routed arrow carry soft
+ * penalties, so arrows keep clearance where there is room but can squeeze through tight gaps.
  */
 public final class ArrowRouter {
```
sbancuz, line 26: If we are using ELK, arrow routing can be done with that, and will probably be more efficient

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
```diff
@@ -456,6 +542,66 @@ private void ensureRoutes() {
         }
 
         edgeRoutes.putAll(ARROW_ROUTER.route(obstacles, requests));
+
+        // The routing input replays headlessly (ArrowRouter is Minecraft-free); dump it at
+        // DEBUG on every recompute so any bad-looking route can be rebuilt from the dev log.
+        // TODO: gate routing diagnostics behind a debug flag for the 1.0 cleanup.
+        if (PlanNH.LOG.isDebugEnabled()) {
```
sbancuz: Add the config, there already is a Config class (I think also in this branch) that you can use

addon.gradle
```diff
@@ -0,0 +1,16 @@
...
+tasks.matching { it.name == 'shadowJar' }.configureEach {
+    exclude 'images/**'
+    exclude 'model/**'
+}
+
+// ELK declares guava as the open range [15.0,), which re-resolves to the newest guava on
+// every fresh resolution - unreproducible builds and a needlessly fat shaded jar. Pin the
+// ELK-0.8-era release. Scoped to the shadow configurations so the dev/runtime classpath
+// (which must keep Minecraft's guava) is untouched.
+configurations.matching { it.name.toLowerCase(Locale.ROOT).contains('shadow') }.configureEach {
```
sbancuz, line 29: Check out here: https://github.com/GTNewHorizons/GuideNH/blob/0004d30077d794bdedae648d98dd8ced687799ba/build.gradle.kts#L19

When implementing elk into guideNH I had to shadow a bunch more stuff

src/main/java/com/sbancuz/plannh/layout/AutoLayout.java
```diff
@@ -0,0 +1,289 @@
...
+
+    /**
+     * Computes new world-space top-left positions for every node. Positions are relative to an
+     * arbitrary origin; callers anchor the result wherever they want.
+     */
+    public static Map<UUID, int[]> layout(final List<NodeBox> nodes, final List<Link> links) {
+        final Map<UUID, int[]> result = new HashMap<>();
+        if (nodes.isEmpty()) return result;
+
+        final List<NodeBox> ordered = canonicalOrder(nodes, links);
```
sbancuz: Why do you need to order them? (I don't know, it's not a critique)

src/main/java/com/sbancuz/plannh/gui/RecipeNodeWidget.java
```diff
@@ -190,7 +188,12 @@ private int calcInfoHeight() {
         return lines * LINE_H + 6;
     }
 
-    private void ensureRecipeHandler() {
+    /**
+     * Loads the NEI handler that determines this widget's real size. Lazy - normally first
+     * draw does it, but MUI2 culls off-viewport widgets, so anything measuring node sizes
+     * (auto-layout) must call this first or off-screen nodes report stub dimensions.
+     */
+    void ensureRecipeHandler() {
```
sbancuz: either private or public

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
```diff
@@ -325,6 +325,92 @@ private void addNodeWidget(final Node node) {
         child(widget);
     }
 
+    /** Pre-auto-layout node positions, so a layout can be undone with shift-click. */
+    @Nullable
+    private Map<UUID, int[]> layoutSnapshot;
```
sbancuz: I feel like this is the wrong approach (for now it can work, but it's a bit too narrow in scope), but eventually we would like to implement undo/redo so it would be completely thrown out when we do.

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
```diff
@@ -325,6 +325,92 @@ private void addNodeWidget(final Node node) {
...
+        final List<AutoLayout.NodeBox> boxes = new ArrayList<>();
+        int anchorX = Integer.MAX_VALUE;
+        int anchorY = Integer.MAX_VALUE;
+        for (final Node node : graph.getNodes()) {
+            final RecipeNodeWidget widget = nodeWidgets.get(node.id);
+            if (widget != null) widget.ensureRecipeHandler();
+            final int width = widget != null ? widget.getWorldWidth() : 120;
+            final int height = widget != null ? widget.getWorldHeight() : 80;
+            boxes.add(
+                new AutoLayout.NodeBox(
```
sbancuz: Why creating new types for the same things?

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
```diff
@@ -325,6 +325,92 @@ private void addNodeWidget(final Node node) {
...
+                    node.machineName == null ? "" : node.machineName,
+                    width,
+                    height,
+                    node.inputs.size(),
+                    node.outputs.size()));
+            anchorX = Math.min(anchorX, node.x);
+            anchorY = Math.min(anchorY, node.y);
+        }
+
+        final List<AutoLayout.Link> links = new ArrayList<>();
```
sbancuz: Same here, this will just cost a lot of allocations

sbancuz (changes requested):
I think fable will hate me after this, also check out: https://github.com/GTNewHorizons/GuideNH/blob/0004d30077d794bdedae648d98dd8ced687799ba/src/main/java/com/hfstudio/guidenh/guide/internal/mermaid/flowchart/ElkLayoutStrategy.java#L31

This will remove the freeze the first time you layout with ELK

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
OrderedSet86: What's your suggestion here? I could try implementing a wider undo/redo, but the scope may be big enough that it should be another PR. I am trying to keep the PR sizes down (hence 3 PRs instead of 1 enormous one) to make them easier to review

sbancuz: Fair, maybe remove it and we can do it later properly

src/main/java/com/sbancuz/plannh/layout/AutoLayout.java
OrderedSet86: I'm going to subject you to direct LLM output here because it explains it quite well. The short version is "to make it invariant to node add order" and the method is "group by SCC and then do topological sort on the SCC groups to make a DAG."

As for why you would want node add order layout invariance, it avoids catastrophic auto-layout shifts with seemingly minor changes, and gives us one answer always for the same problem (which makes reported bugs easier for us devs to solve).

------------------------------------------

The sort is doing two jobs: making the layout **deterministic** (a function of the graph, not of the order recipes were clicked in NEI), and feeding ELK a **flow-aligned traversal order** so its cycle breaker cuts recycle loops in the right place.

### 1. ELK is only deterministic *given* a node order

We pin `RANDOM_SEED`, so ELK itself has no randomness left — but several of its phases consume nodes in **model order** (the order they were added to the `ElkNode`), and break ties by it. Without the sort, model order = NEI click order.

The clearest case is the cycle breaker we use (`CycleBreakingStrategy.DEPTH_FIRST`). From ELK 0.8.1's [`DepthFirstCycleBreaker`](https://github.com/eclipse-elk/elk/blob/master/plugins/org.eclipse.elk.alg.layered/src/org/eclipse/elk/alg/layered/p1cycles/DepthFirstCycleBreaker.java):

```java
for (LNode node : nodes) {                  // <- model order
    if (Iterables.isEmpty(node.getIncomingEdges())) {
        sources.add(node);
    }
}
for (LNode source : sources) {
    dfs(source);
}
// Start more DFS runs for all nodes that have not been visited yet.
// These must be part of a cycle since they are not source nodes
for (int i = 0; i < nodeCount; i++) {       // <- model order again
    if (!visited[i]) {
        dfs(nodes.get(i));
    }
}
```

Any edge the DFS finds pointing back into its current path gets reversed for layering. So for a **source-less loop** (very common here — self-sustaining recycle chains where every machine has an input), *which node happens to come first in model order is the DFS entry point, and that decides which edge gets reversed, which decides the entire left-to-right layering downstream.*

Concretely, take a loop `Electrolyzer → Reactor → Centrifuge → Electrolyzer` with a tap `Centrifuge → Chem Bath`. Every node has an incoming edge, so there are no sources and the DFS entry is pure model order:

```mermaid
flowchart LR
  subgraph A ["User clicked Reactor first → chart reads Reactor → Centrifuge → Electrolyzer"]
    R1[Reactor] --> C1[Centrifuge] --> E1[Electrolyzer]
    E1 -. reversed .-> R1
    C1 --> O1[Chem Bath]
  end
```

```mermaid
flowchart LR
  subgraph B ["User clicked Electrolyzer first → chart reads Electrolyzer → Reactor → Centrifuge"]
    E2[Electrolyzer] --> R2[Reactor] --> C2[Centrifuge]
    C2 -. reversed .-> E2
    C2 --> O2[Chem Bath]
  end
```

Same graph, two click histories, two different pictures. Worse: model order is just insertion order, so deleting and re-adding an unrelated node can reshuffle the whole chart on the next auto-layout. With the canonical sort, auto-layout is idempotent and identical for everyone who builds the same chart.

### 2. Why not something simpler, like sorting by UUID?

That would already make layout deterministic *for one saved chart* — but node UUIDs are random at creation, so the same logical chart rebuilt from scratch would sort differently and lay out differently. And it does nothing for the quality issue below. So the sort key has to come from the *structure* of the graph plus stable machine names; the id is only the last-resort tiebreak between identical machines at identical depth.

### 3. Why SCC-condensation height specifically

The order we actually want is "upstream machines first" — topological order. Then:

- ELK's DFS starts at true sources and walks *with* the flow, so the only edges it reverses are genuine recycle edges (instead of reversing a main-flow edge because the DFS happened to enter a loop from the side);
- on big graphs (>300 nodes) where we enable `CONSIDER_MODEL_ORDER`, ELK's model-order tie-breaking in layering/crossing minimization agrees with flow direction instead of click history;
- height 0 first = sources leftmost, matching `Direction.RIGHT`.

But a topological order doesn't exist for cyclic graphs, and recycle loops are the norm in recipe charts. The standard fix ([cp-algorithms has a good writeup](https://cp-algorithms.com/graph/strongly-connected-components.html)) is to contract each **strongly connected component** into one super-node. The resulting *condensation* is guaranteed to be a DAG, so it *can* be ordered — we take longest-path height from the sources:

```mermaid
flowchart LR
  subgraph S0 ["SCC 0 · height 0"]
    M[Miner]
  end
  subgraph S1 ["SCC 1 · height 1 — recycle loop"]
    Ma[Macerator] --> W[Ore Washer] --> Ce[Centrifuge] --> Ma
  end
  subgraph S2 ["SCC 2 · height 2"]
    EBF[Blast Furnace]
  end
  subgraph S3 ["SCC 3 · height 3"]
    As[Assembler]
  end
  M --> Ma
  Ce --> EBF
  EBF --> As
  style S0 fill:#d8f5d0,color:#000
  style S1 fill:#fbd2d2,color:#000
  style S2 fill:#fdf3c0,color:#000
  style S3 fill:#d3dcfb,color:#000
```

condenses to:

```mermaid
flowchart LR
  A0["{Miner}<br/>h = 0"] --> A1["{Macerator, Ore Washer, Centrifuge}<br/>h = 1"] --> A2["{Blast Furnace}<br/>h = 2"] --> A3["{Assembler}<br/>h = 3"]
  style A0 fill:#d8f5d0,color:#000
  style A1 fill:#fbd2d2,color:#000
  style A2 fill:#fdf3c0,color:#000
  style A3 fill:#d3dcfb,color:#000
```

Sort key is `(SCC height, name, id)`. All members of a loop share one height, so within a loop the order falls through to the name — which is exactly what makes the DFS entry point into each loop (case 1 above) deterministic too: it's always the alphabetically first machine in the loop, not whichever one was clicked first.

Tarjan is just the linear-time way to get the SCCs; it's written iteratively because recursion depth would be the chart's path length.

OrderedSet86:
need to warm up ELK engine first to mitigate lag

sbancuz:
> need to warm up ELK engine first to mitigate lag

Yes, that is the idea (it only mitigates the first lag spike though, nothing major). Since we don't have the mechanism that I linked in guideNH check out the commits for the background + shader. There I use a function to load stuff (hopefully) asynchronously during mod loading

src/main/java/com/sbancuz/plannh/layout/AutoLayout.java
OrderedSet86: One more addition to avoid layout shifts. Now graph coloring is used to guarantee a unique solution based on graph metadata (neighbors etc). Also edges follow a canonical order as well.

--------------------------------------

### 4. Update: closing the last gap — the id tiebreak

Section 3 left one honest hole: the final id tiebreak. Ids are random at creation, so two machines tying on `(height, name)` still ordered randomly — the same logical chart rebuilt from scratch could swap them. The commit closes this with **wiring colors**: an iterated graph-coloring pass (color refinement) that identifies each node by its surroundings instead of its random id.

**Where the id used to leak through.** One Miner output feeds two Chemical Reactors (same machine, same size), one line continuing to a Blast Furnace, the other to a Distillery. Both reactors tie on `(height, name)`, and because both edges leave the *same output port*, both vertical stackings have equal edge crossings — a genuine tie that only model order breaks. So whichever reactor happened to draw the smaller id decided the picture; in a headless harness that rebuilds the chart with fresh random ids and shuffled insertion order, 17 of 39 rebuilds flipped it:

```mermaid
flowchart LR
  subgraph A ["Rebuild 1 — EBF branch happens to sort first"]
    M1[Miner] --> RA1[Chemical Reactor] --> E1[Blast Furnace]
    M1 --> RB1[Chemical Reactor] --> S1[Distillery]
  end
```

```mermaid
flowchart LR
  subgraph B ["Rebuild 2 — Distillery branch happens to sort first"]
    M2[Miner] --> RB2[Chemical Reactor] --> S2[Distillery]
    M2 --> RA2[Chemical Reactor] --> E2[Blast Furnace]
  end
```

**Links leak the same way.** Holding the nodes, the ids and the node order all fixed and shuffling *only* the order links were drawn in still flips the branches on 19 of 39 rebuilds — node positions, not just edge routes: with equal crossings and fixed ports, ELK keeps whichever order it encounters first, and encounter order follows the edge list. So links are also sorted to follow the node canonicalization — `(source position, output port, target position, input port)` — which keeps edge routing stable for free.

**The coloring.** To pin the node order down, every node gets a canonical color. Take a harder version of the chart above, where the twin branches stay identical for one more hop before diverging.

Round 0 — color nodes by what they *are*: `(size, port counts, height, name)`. All four Chemical Reactors are indistinguishable and share a color:

```mermaid
flowchart LR
  M0[Miner] --> A0[Chemical Reactor] --> CA0[Chemical Reactor] --> E0[Blast Furnace]
  M0 --> B0[Chemical Reactor] --> CB0[Chemical Reactor] --> S0[Distillery]
  style M0 fill:#d8f5d0,color:#000
  style A0 fill:#fbd2d2,color:#000
  style B0 fill:#fbd2d2,color:#000
  style CA0 fill:#fbd2d2,color:#000
  style CB0 fill:#fbd2d2,color:#000
  style E0 fill:#d3dcfb,color:#000
  style S0 fill:#fdf3c0,color:#000
```

Round 1 — recolor every node by its own color plus the sorted list of `(own port, direction, peer color)`. The two middle reactors split, because one's output peer is a Blast Furnace and the other's is a Distillery. The two left reactors still tie: both see a green Miner upstream and a red reactor downstream:

```mermaid
flowchart LR
  M3[Miner] --> A3[Chemical Reactor] --> CA3[Chemical Reactor] --> E3[Blast Furnace]
  M3 --> B3[Chemical Reactor] --> CB3[Chemical Reactor] --> S3[Distillery]
  style M3 fill:#d8f5d0,color:#000
  style A3 fill:#fbd2d2,color:#000
  style B3 fill:#fbd2d2,color:#000
  style CA3 fill:#e8d5f5,color:#000
  style CB3 fill:#fde3c8,color:#000
  style E3 fill:#d3dcfb,color:#000
  style S3 fill:#fdf3c0,color:#000
```

Round 2 — recolor again, now with the round-1 colors. The left reactors split too, because their downstream peers no longer match:

```mermaid
flowchart LR
  M4[Miner] --> A4[Chemical Reactor] --> CA4[Chemical Reactor] --> E4[Blast Furnace]
  M4 --> B4[Chemical Reactor] --> CB4[Chemical Reactor] --> S4[Distillery]
  style M4 fill:#d8f5d0,color:#000
  style A4 fill:#cdeef0,color:#000
  style B4 fill:#f5c6dc,color:#000
  style CA4 fill:#e8d5f5,color:#000
  style CB4 fill:#fde3c8,color:#000
  style E4 fill:#d3dcfb,color:#000
  style S4 fill:#fdf3c0,color:#000
```

Every node now has its own color, so one more round changes nothing and the refinement stops. In general, a difference k hops away separates twins after k rounds — which is why the pass iterates at all: a single round would have stopped after the middle diagram, left the two red reactors tied on the random id, and the harness shows exactly that chart flipping on 19 of 39 rebuilds.

Two details that make the colors canonical: within a round, each new color is the *rank of the node's signature in the sorted list of distinct signatures*, so colors depend only on the graph, never on iteration order; and the signature includes port indices, so "feeds input 0" and "feeds input 1" of the same peer are different wirings.

The sort key becomes `(height, name, color, id)`. After this change, the id correctly orders all but true structural twins — and with iteration, "twin" now means *no chain of links anywhere in the chart distinguishes the two nodes*, so swapping them is invisible on screen.

**Cost.** Each round builds signatures in O(V + E) and ranks them with one sort; every round before the fixpoint must split at least one color class, so rounds ≤ V, bounding the worst case by O(V·(V+E)·log V). In practice rounds = distance to the nearest distinguishing feature, 2–4 on real charts. As a deliberately adversarial measurement: a 500-node chain of identically named machines (forces exactly one hop of progress per round, ~500 rounds) runs the whole `layout()` call, ELK included, in ~310 ms.

**Results.** Three harness scenarios — same-port fan-out, disconnected same-name components, and the two-hop twins above — each rebuilt 40 times with fresh random ids and shuffled node/link insertion: all coordinate-identical. The existing reference charts lay out unchanged.

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
OrderedSet86: Fable preferred to do this because it made it Minecraft-free (i.e. headless outside the game) but this doesn't matter because these kind of tests are going to be handled by JUnit in the future. So I agree, moving the actual canvas nodes is the preferred method

src/main/java/com/sbancuz/plannh/gui/ArrowRouter.java
OrderedSet86: ArrowRouting is done during interactive dragging, ELK running every frame would be tremendously laggy. So do you mean remove interactive dragging and only update after the node is placed? That seems like a UX downgrade

OrderedSet86: > ELK running every frame would be tremendously laggy

It occured to me that maybe we could only pull out the section of ELK edge router that works on the edges connected to the node being moved, which would make it performant enough, but that won't work (LLM output):

-------------------------------------

I checked whether ELK's edge routing can be pulled out and run standalone on fixed node positions; it can't:

- `OrthogonalEdgeRouter`'s precondition is "the graph has a proper layering with assigned node and port positions" — it routes only in the gaps between adjacent layers of the internal `LGraph`, with long edges pre-split by dummy nodes that earlier pipeline phases insert. After a free-form drag that structure doesn't exist.
- Its postcondition is "each node is assigned a horizontal coordinate" — the routing phase sets layer x-positions as it reserves edge channels, so it moves nodes; it would fight the drag.
- The phases are a processor list on the private `LGraph`; the only partial-pipeline API is the layout unit-test harness, which still runs cycle breaking → layering → crossing minimization → placement first, i.e. a full layout.
- ELK's own answer for "route around fixed nodes" is `org.eclipse.elk.alg.libavoid` — and that Eclipse wrapped a C++ library rather than reusing their Java router says there's nothing extractable. It's also not JNI: `LibavoidServer` spawns a separate `libavoid-server` executable and talks to it over stdin/stdout pipes with a watchdog. Eclipse doesn't distribute the binaries (the repo's `libavoid-server/` holds only LICENSE files — you build them from C++ per platform, and only linux64/win64/osx-x64 are supported, no ARM macs). And for the drag use-case specifically the economics invert: a recompute per drag frame through serialize→pipe→native-process→parse is slower than the in-process grid A* we have — ELK keeps a LibavoidServerPool precisely because those calls are expensive.

So the division of labor here mirrors ELK's own ecosystem: layout engine for arranging (ELK layered), standalone connector router for fixed-position routing (libavoid there, ArrowRouter here — a pure-Java, in-process, Minecraft-free stand-in for the same role).

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
OrderedSet86: see other allocation comment

OrderedSet86:
Warm start is now called in `ClientProxy.init`

I added a new step to the ordering algorithm that should handle all remaining edge cases - now the nodes are uniquely ID'd by a graph coloring algorithm (Weisfeiler–Lehman).

I placed the full technical explanation into the root PR description, with toggles for the long sections

All feedback should hopefully be addressed now

src/main/java/com/sbancuz/plannh/ClientProxy.java
```diff
@@ -54,6 +55,13 @@ public void init(final FMLInitializationEvent event) {
         FMLCommonHandler.instance()
             .bus()
             .register(this);
+
+        // ELK's first layout pays for its class loading and metadata registration; warm it up
+        // off-thread so the first Auto-Layout click doesn't freeze the game.
+        final Thread elkWarmup = new Thread(AutoLayout::warmup, "PlanNH ELK warmup");
```
sbancuz: Try to not create new threads

sbancuz (changes requested):

OrderedSet86:
^ added correctness bugfixes

sbancuz (approved):
LGTM

PR #20, opened 2026-07-20
Small QOL fixes

OrderedSet86:
1. Patch MUI logspam: `Widget Row is larger with padding on axis Y than parent Row. Padding can't be applied correctly!` (NOTE: this is currently duplicated in auto-layout, I will remove it soon to avoid merge conflict)
2. Add a repeat functionality if the -/+ button is held down in the config window. This is handly for making integer throughputs for factory chart demos
3. Fix an edge case where the buttons could desync with the canvas state. I noticed this when deleting a chart slot in Auto mode (a new addition from my factory solver branch) - it would keep the button in AUT, but it would silently switch to OUT under the hood (which was visible in the summary node).
4. Not sure if this actually happens in normal practice, but I had the game crash with GT recipe nulls when I was importing platline's gtnh-flow YAML chart. (This may be when it was trying to import non-supported machine types.) This PR adds a defensive handling of GT recipe input/outputs being null.

### DIFF
--- a/src/main/java/com/sbancuz/plannh/data/provider/gregtech/GTProvider.java
+++ b/src/main/java/com/sbancuz/plannh/data/provider/gregtech/GTProvider.java
@@ -275,8 +275,9 @@ public class GTProvider implements PropertyProvider {
         node.inputs.clear();
         node.outputs.clear();
 
+        // GT recipe arrays may contain null slots (e.g. gap outputs); skip them.
         for (int i = 0; i < r.mInputs.length; i++) {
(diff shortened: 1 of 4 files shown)


sbancuz (approved):
I mean, I can approve, but I feel like 3/4 commits will be yeeted, the only one salvageable is the null checks for GTRecipes, but when I added them before it seemed to work without crashing so idk

OrderedSet86:
Fine by me, these will be handy to have in the near term (next couple weeks)

PR #21, opened 2026-07-20
Add headless test harness with gtnh-flow YAML bootstrap tests

OrderedSet86:
NOTE: the test harness is used more extensively later on in my upcoming factory solver PR, but I thought it would be a good idea to break that one up into multiple smaller PRs so they could be addressed individually. Hence this PR

## Changes
1. Add JUnit for testing
2. Add snakeyaml to load gtnh-flow YAML files
3. Add gtnh-flow YAML files for bootstrapping tests
4. Add basic solver tests (enumerated below)

## Tests (ported from my flowv2 testing repo)
1. Smoke test: try balancing all gtnh-flow YAML charts. Must satisfy:
    - Doesn't crash
    - Doesn't take longer than 15s, including on nanocircuits (which is appreciable fraction of stargate)
    - Must return a solved result
2. Solver output follows basic sanity checks, such as machine count, connectedness, etc. (This is primarily a check of the gtnh-flow import compat - these wouldn't be expected to be violated in normal conditions)
3. Summary node follows basic sanity checks, i.e. summing all machine nodes equals the summary node


src/test/java/com/sbancuz/plannh/harness/ConservationValidator.java
```diff
@@ -0,0 +1,62 @@
...
+import com.sbancuz.plannh.data.flowchart.Balancer.NodeBalance;
+import com.sbancuz.plannh.data.flowchart.Node;
+
+/**
+ * Independent aggregation over a balance result: per-ingredient production and consumption,
+ * summed across every port and normalized to the balance's common cycle. This is the same
+ * accounting the summary widget displays, recomputed from the raw node balances, so tests can
+ * assert the summary against it without trusting the summary's own code.
+ */
+public final class ConservationValidator {
```
sbancuz: Ehm.. why have 2 of the same stuff? I get you are trying to test the summary, but if the calculations are the same I don't see the point

src/test/java/com/sbancuz/plannh/harness/GtnhFlowLoader.java
```diff
@@ -0,0 +1,177 @@
...
+                    List.of(Settings.MACHINES.def(), Settings.TICK_MODIFIER.def()),
+                    (s, ctx) -> new MachineProfile.EffectResult(ctx.recipeDuration(), 0, 1)));
+        }
+    }
+
+    private GtnhFlowLoader() {}
+
+    /** Loads {@code /gtnh-flow/<name>.yaml} from test resources. */
+    public static LoadedChart load(final String name) {
+        final InputStream in = GtnhFlowLoader.class.getResourceAsStream("/gtnh-flow/" + name + ".yaml");
```
sbancuz: I've never seen this pattern before, not sure what are the implications. Also this way it seems quite limiting, you can't import it from anywhere else

src/test/java/com/sbancuz/plannh/harness/GtnhFlowLoader.java
```diff
@@ -0,0 +1,177 @@
...
+ * <ul>
+ * <li>{@code number: N} - the machine count is fixed to N (maps to PlanNH's fixed machine
+ * count).</li>
+ * <li>{@code target: {ingredient: rate}} - a desired output rate; PlanNH has no equivalent yet,
+ * so it is surfaced on the {@link LoadedChart} for the solver work to consume.</li>
+ * </ul>
+ */
+public final class GtnhFlowLoader {
+
+    public record Pin(String kind, int machineIndex, String machineName, String ingredient, double value) {}
```
sbancuz, line 44: What's a pin?

sbancuz (changes requested):
I like the idea, but I'm not too sure about the implementation, especially the ConservationValidator. Also feels a bit weird to have the plans of another project as tests, but I guess that's fine

OrderedSet86:
> Also feels a bit weird to have the plans of another project as tests, but I guess that's fine

Understandable. This was meant for bootstrapping so I didn't have to draw out platline etc. again. I spent many hours making machine lines for gtnh-flow (as you might be able to gather from the YAML sizes). I am fine with replacing them later with the export format for PlanNH, but this lets the big solver PR come sooner, which is the part that will hopefully improve end user experience

src/test/java/com/sbancuz/plannh/harness/ConservationValidator.java
OrderedSet86: Yeah this test will be dropped

src/test/java/com/sbancuz/plannh/harness/GtnhFlowLoader.java
OrderedSet86: This is the "target" arg in gtnh-flow - i.e. user saying "this machine must produce 1000 chlorine/s"

It's a placeholder, this eventually needs to be put into the wider PlanNH solver

sbancuz: In the meantime you could approximate the amount by changing the MACH setting until you arrive at the target

OrderedSet86:
Nanocircuits exceeds the budget now, apparently pin wasn't being applied properly before to fix the machine count. Now that it does nanocircuits takes 19s. I think it makes sense to leave the test in a failing state, it's something that needs to get fixed later

src/test/java/com/sbancuz/plannh/harness/GtnhFlowLoader.java
OrderedSet86: Apparently this is a standard JUnit way of loading resources. LLM explainer output:

-------------------------------------------------------------

- On the pattern itself: `getResourceAsStream` is the standard JUnit fixture idiom — gradle puts `src/test/resources` on the test classpath, so the lookup resolves identically under `gradlew test`, IDE runners and CI, and keeps working if tests ever run from a jar. Its two real quirks are that it returns `null` instead of throwing when the resource is missing (guarded, with the chart name in the message), and that classloader-sensitive environments like Forge's LaunchClassLoader can complicate resource lookup — but that doesn't apply here since these tests run on the plain JVM classpath, no Forge classloader involved. There's a comment at the lookup now covering this.
- On "can't import it from anywhere else" — agreed, that was a real limitation, and it's fixed structurally: the general-purpose entry point is `load(String name, InputStream in)`, so charts can be fed from any file, another project, or a future PlanNH export (`name` seeds the deterministic node/edge ids). The classpath `load(name)` is just the corpus convenience, delegating to the stream entry point.
- The `static { ensureDefaultMachineProfile(); }` class-init side effect is gone; `load` calls the (idempotent) registration itself. It exists because headless tests need the default machine profile that mod init normally registers, and `MachineProfile.Builder` reads `NEIClientConfig` in its constructor, which needs a running client — that's documented on the method.

OrderedSet86:
ok, now done :)

sbancuz:
> Nanocircuits exceeds the budget now, apparently pin wasn't being applied properly before to fix the machine count. Now that it does nanocircuits takes 19s. I think it makes sense to leave the test in a failing state, it's something that needs to get fixed later

We can't have failing tests

OrderedSet86:
> > Nanocircuits exceeds the budget now, apparently pin wasn't being applied properly before to fix the machine count. Now that it does nanocircuits takes 19s. I think it makes sense to leave the test in a failing state, it's something that needs to get fixed later
> 
> We can't have failing tests

Nanocircuit generation time pass (like 10x the size of platline) isn't on the beta (performant half-platline) or maybe even 1.0 release list, so it's not worth the effort to try and make it pass right now, but it's a good test to have handy. Maybe I can mark it to not run?

sbancuz:
> > > Nanocircuits exceeds the budget now, apparently pin wasn't being applied properly before to fix the machine count. Now that it does nanocircuits takes 19s. I think it makes sense to leave the test in a failing state, it's something that needs to get fixed later
> > 
> > 
> > We can't have failing tests
> 
> Nanocircuit generation time pass (like 10x the size of platline) isn't on the beta (performant half-platline) or maybe even 1.0 release list, so it's not worth the effort to try and make it pass right now, but it's a good test to have handy. Maybe I can mark it to not run?

Sure, or add time. What I need test that pass because otherwise CI would fail (here there is no CI to run so you don't see it)

sbancuz (approved):
LGTM

PR #22, opened 2026-07-22
Add stack-based undo/redo using graph snapshots

OrderedSet86:
Supports keybinds:
Undo: Ctrl+Z
Redo: Ctrl+Shift+Z or Ctrl+Y

Keybinds route through NEI so they are configurable by end user

### DIFF
--- a/src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
+++ b/src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
@@ -6,8 +6,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
+import net.minecraft.client.gui.GuiScreen;
+
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.lwjgl.input.Keyboard;
 import org.lwjgl.opengl.GL11;
 
 import com.cleanroommc.modularui.api.UpOrDown;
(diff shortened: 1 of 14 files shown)


src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
```diff
@@ -886,6 +938,18 @@ public void onGroupEditingDone() {
 
     @Override
     public @NotNull Result onKeyPressed(final char typedChar, final int keyCode) {
+        if (GuiScreen.isCtrlKeyDown() && keyCode == Keyboard.KEY_Z) {
```
sbancuz: Idk if it's possible to use configurable keybinds, maybe through NEI configs

sbancuz (changes requested):

src/main/java/com/sbancuz/plannh/gui/CanvasWidget.java
OrderedSet86: done :)

OrderedSet86:
- Code quality improved through the use of `PlanAPI.recordEdit` for graph mutations. This allows multiple operations to clearly count as one atomic operation for the purposes of undo/redo
- Use NEI keybind + in-game config for undo/redo

OrderedSet86:
LLM (caveman) generated list of deferred tasks:

**PR Candidates**
- Missing lang keys for the existing F8 bind: ClientProxy registers key.neiflowchart.open / key.categories.neiflowchart, neither is in en_US.lang, so vanilla Controls shows the raw key string. Two lines, and the PR already touches that file for the NEI binds. Cheapest thing on this list.
- Undo pressed mid-drag leaves a stale dragEditToken; the drag's onDragEnd then commits against the swapped graph. Same class as the focus bug just fixed, but needs a hand on the keyboard while a mouse button is held. A dragEditToken = null on restore would cover it.

**Future Changes**
- Balancer.java:241 writes machineConfig.setMachineCount(...) during render-time solving, so serialized state mutates outside every bracket. Undo restores stale counts, and solver drift can defeat commitEdit's no-op check. The fix is "stop persisting derived state" — fixed on sbancuz branch
- Snapshot cost: Gson with pretty-printing + full ingredient NBT per node, two encodes per drag, MAX_DEPTH 64 on both stacks; decode re-runs RecipeHandlerRef.of against a 200-entry LRU, so undo on a large graph can fire hundreds of synchronous NEI lookups. Worth stating honestly given the large-graphs branch — a compact snapshot codec is its own PR.

Minor, deliberately left: deleteSlot isn't undoable (slot and its history die together — consistent with per-slot history); FlowchartScreen.graph goes stale after a restore but has no readers.

sbancuz (approved):
LGTM

PR #25, opened 2026-07-29
Add AUTO mode flow solver

OrderedSet86:
## Features

- Add AUTO, a new MILP solver mode. Uses fractional machine counts. Solves tough cases from https://github.com/OrderedSet86/flow2/blob/main/math.md and generally seems to perform well on all the hand charts I checked. Uses heuristics to give a maximally reasonable default. Does not solve until user pins a node (either "fixed" or new target args)
- Add target UI in config for pinning an edge quantity (text entry, focus on open)
- Add diagnostic output (controlled by solver) in summary
- Upgrade ojAlgo 47.3.1 -> 57.1.0 (2x faster and less buggy)
- Set AUTO to default solver

## Correctness

- Handle crashes caused by malformed graphs without world panic
- Divide by 10k instead of 100 for chance items (GT5u ranges from 1-10000, where 10000 = 100%)

## Tests

- AUTO mode solver maximum time tests (similar to layout)
- Ground truth tests (known quantities from hand solving that AUTO solver must come up with)
- Tiebreaker test (`mk1_exactlyOneGate_sinkExcessPreferred`) - sinks should be preferred in AUTO
- Make sure machines reachable by a pin always run (in AUTO)
- Various tests to ensure pins and disconnected nodes work properly (in AUTO)
- Various tests for diagnostic output (for AUTO)
- Add some more gtnh-flow YAMLs

## Example AUTO solutions

1. Partial sulfuric acid recycling
image

2. Parallel ratio differences. There are multiple valid solutions here but it picks the most reasonable one, which is excess heavy naquadah. I made it output said heavy naquadah to make Tiberium, and the heavy naquadah fuel ratio between the fusion reactor and the chemical bath is the expected 12:1.
image

3. 1000L/s Cetane-Boosted Fuel
image

4. 40L/s RP-1 Rocket Fuel. This one is a little weird since it chose charcoal as its excess product; a more natural excess in reality would be excess nitrogen. But this one of those "many right answers" problems
image

5. Palladium line (coming soon)

### DIFF
--- a/src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
+++ b/src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
@@ -1,0 +1,1062 @@
+package com.sbancuz.plannh.data.flowchart;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+
+import org.ojalgo.optimisation.Expression;
+import org.ojalgo.optimisation.ExpressionsBasedModel;
+import org.ojalgo.optimisation.Optimisation;
+import org.ojalgo.optimisation.Variable;
+
+import com.sbancuz.plannh.data.MachineConfig;
+
+/**
+ * Four-stage lexicographic MILP balancer over PlanNH's machine-node graphs, in recipe-extent
+ * form. The chart plus its pins is the entire input:
+ * no pin means no balancing (an unpinned chart is just wiring - {@link #NO_PIN}); one or more
+ * pins anchor the scale and everything else - including where external sources/sinks go - is
+ * decided automatically, never demanded from the user.
+ *
+ * <p>
+ * Model: one extent variable per machine (crafts/second), one flow variable per drawn edge
+ * (ingredient/second). Per connected port: {@code sum(edge flows) + external = extent *
+ * perCraftQty}. Unconnected ports are free terminals (supplying oil / collecting product is the
+ * point of the chart, not a cost). Connected ports get an external whose gate is shared per
+ * ingredient and direction: all input-side externals of one ingredient open one SOURCE gate, all
+ * output-side externals one SINK gate (ingredients are identified structurally, as the
+ * connected components of ports under the drawn edges). Ingredient-level gating keeps the
+ * binary count ~2x intermediates instead of one per port.
+ *
+ * <p>
+ * Stages, each optimized subject to the previous optima:
+ * <ol>
+ * <li>Minimize the number of open gates (sinks weigh 1024, sources 1025, so equal counts prefer
+ * "discard the excess" over "supply an intermediate from outside").</li>
+ * <li>Minimize total external quantity, weighted gate count capped at stage-1's optimum.</li>
+ * <li>Minimize total internal flow, gates fixed to stage-2's support (derived from FLOWS, never
+ * binary values - big-M x integrality tolerance leaks otherwise). Pins the free circulation of
+ * fully-recycling loops. Supports tied at (count, quantity) are each tried and the least
+ * internal flow wins, deterministically.</li>
+ * </ol>
+ * Stage 0 ("every machine runs") is implemented as scale-relative extent floors applied in a
+ * second pass only, when the floor-free solution left machines idle - a fixed floor can exceed a
+ * machine's natural rate and conjure phantom externals (the light_fuel trap).
+ *
+ * <p>
+ * A zero-gate LP fast path (all externals closed, minimize internal flow) covers charts that
+ * balance without externals in a single solve. Every accepted solution is validated
+ * independently against the port-conservation rows; a solution that fails validation is rejected
+ * rather than returned (never trust solver status codes).
+ */
+public final class AutoBalancer {
+
+    private static final double DEFAULT_BIG_M = 1e6;
+    private static final int MAX_M_GROWTHS = 3;
+    private static final long STAGE_TIME_LIMIT_MILLIS = 15_000;
+    /**
+     * Budget for the exact stage-1 MILP once the LP deletion filter has already produced a
+     * minimal support. Big-M count minimization gives branch-and-bound no usable root bound, so
+     * on large charts ojAlgo cannot prove optimality in any budget - the filter's answer is used
+     * and marked uncertified instead of burning the full stage budget.
+     */
+    private static final long MILP_CERT_BUDGET_MILLIS = 5_000;
+    /** Flows below this count as zero when deriving gate support. */
+    private static final double ZERO = 1e-6;
+    /** A machine "runs" if its crafts/s exceeds this. */
+    private static final double USE_EPS_DETECT = 1e-7;
+    /** Fallback pass-2 floor (crafts/s) when no machine ran at all. */
+    private static final double USE_EPS = 1e-4;
+    /** Relative slack on the stage-2 quantity cap in stage 3. */
+    private static final double QTY_EPS = 1e-7;
+    /** Relative residual tolerance for independent solution validation. */
+    private static final double VALIDATE_TOL = 1e-6;
+    private static final double SNK_WEIGHT = 1024.0;
+    private static final double SRC_WEIGHT = 1025.0;
+    private static final int TICKS_PER_SECOND = 20;
+    private static final int MAX_ENUMERATED_ALTERNATIVES = 10;
+    private static final int MAX_TIED_SUPPORTS = 5;
+
+    private AutoBalancer() {}
+
+    /**
+     * The failure reason when no machine is pinned. Not an error: an unpinned chart is just
+     * wiring (the gtnh-flow contract) - the model is homogeneous and every solution scales
+     * freely, so any numbers would be an invented anchor the user never asked for.
+     */
+    public static final String NO_PIN = "nothing is pinned - fix a machine count to ask for a balance";
+
+    /** A port on a specific machine. {@code input} distinguishes the two port lists. */
+    public record PortRef(UUID nodeId, int portIndex, boolean input) {}
+
+    /** An external attached to a port, in ingredient units per second. */
+    public record External(PortRef port, double ratePerSecond) {}
+
+    public record Solution(Map<UUID, Double> extentsPerSecond, Map<UUID, Double> machineCounts,
+        Map<UUID, Double> edgeFlowsPerSecond, List<External> gatedSources, List<External> gatedSinks,
+        List<External> terminalInputs, List<External> terminalOutputs, int openGates, double externalQuantity,
+        double totalInternalFlow, boolean floorsUsed, long wallMillis, List<String> notes) {}
+
+    public record Result(Solution solution, String failure) {
+
+        public boolean isSuccess() {
+            return solution != null;
+        }
+
+        static Result ok(final Solution s) {
+            return new Result(s, null);
+        }
+
+        static Result fail(final String reason) {
+            return new Result(null, reason);
+        }
+    }
+
+    /** Solves using the graph's fixed machine counts as the only pins. */
+    public static Result solve(final Graph graph) {
+        return solve(graph, Map.of());
+    }
+
+    /**
+     * @param extraExtentPins additional extent pins (crafts/second) by node id, e.g. from a
+     *                        target-rate pin: extent = targetRate / perCraftQty.
+     */
+    public static Result solve(final Graph graph, final Map<UUID, Double> extraExtentPins) {
+        final long start = System.currentTimeMillis();
+        final Ctx ctx = new Ctx(graph, extraExtentPins);
+        if (ctx.machines.isEmpty()) {
+            return Result.fail("empty graph");
+        }
+        if (!ctx.anyPin) {
+            return Result.fail(NO_PIN);
+        }
+
+        // Pass 1: floor-free.
+        Attempt attempt = runStages(ctx, null);
+        if (attempt.failure != null) {
+            return Result.fail(attempt.failure);
+        }
+
+        // Stage 0, pass 2: if machines idle, retry with scale-relative floors; keep the honest
+        // pass-1 result if the floored model fails.
+        boolean floorsUsed = false;
+        final double[] floors = ctx.floorsFrom(attempt.extents);
+        if (floors != null) {
+            final Attempt floored = runStages(ctx, floors);
+            if (floored.failure == null) {
+                attempt = floored;
+                floorsUsed = true;
+            }
+        }
+
+        final String residualError = ctx.validate(attempt);
+        if (residualError != null) {
+            return Result.fail("solution failed independent validation: " + residualError);
+        }
+
+        return Result.ok(ctx.toSolution(attempt, floorsUsed, System.currentTimeMillis() - start));
+    }
+
+    /**
+     * Enumerates gate supports tied with the optimum at raw gate COUNT (no-good cuts), so a
+     * GUI can ask the user once instead of guessing. Each entry is the set of ports
+     * whose externals carried flow; element 0 is the deterministic default {@link #solve} picks.
+     * Weighted tiebreaks (sink vs source) do not disqualify an alternative.
+     */
+    public static List<Set<PortRef>> enumerateAlternatives(final Graph graph, final Map<UUID, Double> extraExtentPins) {
+        final Ctx ctx = new Ctx(graph, extraExtentPins);
+        if (ctx.machines.isEmpty() || !ctx.anyPin) return List.of();
+
+        final Attempt base = runStages(ctx, null);
+        if (base.failure != null) return List.of();
+        final double[] floors = ctx.floorsFrom(base.extents);
+        final Attempt attempt = floors == null ? base : orElse(runStages(ctx, floors), base);
+
+        final List<Set<PortRef>> supports = new ArrayList<>();
+        supports.add(ctx.flowPortRefs(attempt.externals));
+        // Alternatives are only enumerable when the optimum was certified: on the filter path
+        // there is no optimality frontier to walk, and each cut solve would burn the MILP budget.
+        if (attempt.support.isEmpty() || !attempt.certified) return supports;
+
+        final int optimumCount = attempt.support.size();
+        final double[] activeFloors = floors != null && attempt != base ? floors : null;
+        final Set<Set<Integer>> seen = new HashSet<>();
+        seen.add(attempt.support);
+        final List<Set<Integer>> cuts = new ArrayList<>(seen);
+        while (supports.size() < MAX_ENUMERATED_ALTERNATIVES) {
+            final StageSolve s1 = solveStage1(ctx, activeFloors, cuts, null);
+            // A repeated flow support means the frontier is exhausted: the cuts forbid the seen
+            // BINARY vectors, so the solver can only reproduce a seen support by opening junk
+            // zero-flow gates - and a junk combination (k+1 gates) always costs more than any
+            // genuine k-gate support would, so genuine alternatives are found first.
+            if (s1 == null || s1.support.size() > optimumCount || !seen.add(s1.support)) break;
+            supports.add(ctx.flowPortRefs(s1.externals));
+            cuts.add(s1.support);
+        }
+        return supports;
+    }
+
+    private static Attempt orElse(final Attempt preferred, final Attempt fallback) {
+        return preferred.failure == null ? preferred : fallback;
+    }
+
+    // ---------------------------------------------------------------------------------------
+    // Stage pipeline
+    // ---------------------------------------------------------------------------------------
+
+    /** One full lexicographic run (stages 1-3) under the given floors (null = floor-free). */
+    private static Attempt runStages(final Ctx ctx, final double[] floors) {
+        // Fast path: if the chart balances with every gate closed, stages 1-2 are trivially
+        // optimal (0 gates, 0 external quantity) and this LP IS stage 3.
+        final StageSolve fast = solveStage3(ctx, floors, Set.of(), 0.0);
+        if (fast != null) {
+            return Attempt.of(fast, Set.of(), true, List.of());
+        }
+
+        // Stage 1 runs twice: an LP deletion filter that always produces a minimal support fast,
+        // then a short exact-MILP slice that certifies (or beats) it when the chart is small
+        // enough for branch-and-bound. Large charts keep the filter answer, uncertified.
+        final Set<Integer> filterSupport = deletionFilter(ctx, floors);
+        if (filterSupport == null) {
+            return Attempt.failed("stage 1 (gate count) found no feasible support");
+        }
+        Set<Integer> s1Support = filterSupport;
+        boolean certified = false;
+        final StageSolve milp = solveStage1(ctx, floors, List.of(), ctx.weightedCost(filterSupport));
+        if (milp != null && ctx.weightedCost(milp.support) <= ctx.weightedCost(filterSupport) + 0.5) {
+            s1Support = milp.support;
+            certified = milp.provenOptimal;
+        }
+        final List<String> notes = certified ? List.of()
+            : List.of(
+                "gate count " + s1Support.size() + " is minimal but not certified optimal (exact search over budget)");
+        final double weightedCap = ctx.weightedCost(s1Support);
+
+        final StageSolve s2 = certified ? solveStage2Cut(ctx, floors, weightedCap, List.of())
+            : solveStage2Fixed(ctx, floors, s1Support);
+        if (s2 == null) {
+            return Attempt.failed("stage 2 (external quantity) found no solution within budget");
+        }
+
+        // Ties at (count, quantity) are resolved by stage 3's objective: re-run stage 3 for each
+        // tied stage-2 support and keep the least internal flow. This is what makes loopGraph
+        // deterministically pick the diluted-acid source over the sulfuric one. Only meaningful
+        // on the certified path (the filter path has no optimality frontier to enumerate).
+        final List<Set<Integer>> candidates = certified ? tiedSupports(ctx, floors, weightedCap, s2)
+            : List.of(s2.support);
+        StageSolve best = null;
+        Set<Integer> bestSupport = null;
+        for (final Set<Integer> support : candidates) {
+            final StageSolve s3 = solveStage3(ctx, floors, support, s2.externalQuantity);
+            if (s3 != null && (best == null || s3.internalFlow < best.internalFlow - ZERO)) {
+                best = s3;
+                bestSupport = support;
+            }
+        }
+        if (best == null) {
+            return Attempt.failed("stage 3 (internal flow) found no solution within budget");
+        }
+        return Attempt.of(best, bestSupport, certified, notes);
+    }
+
+    /** The stage-2 support plus any other supports tied at the same (gate count, quantity). */
+    private static List<Set<Integer>> tiedSupports(final Ctx ctx, final double[] floors, final double weightedCap,
+        final StageSolve s2) {
+        final List<Set<Integer>> candidates = new ArrayList<>();
+        candidates.add(s2.support);
+        final List<Set<Integer>> cuts = new ArrayList<>();
+        cuts.add(s2.support);
+        while (candidates.size() < MAX_TIED_SUPPORTS) {
+            final StageSolve next = solveStage2Cut(ctx, floors, weightedCap, cuts);
+            if (next == null || next.externalQuantity > s2.externalQuantity * (1 + QTY_EPS) + ZERO
+                || next.support.size() > s2.support.size()) {
+                break;
+            }
+            candidates.add(next.support);
+            cuts.add(next.support);
+        }
+        return candidates;
+    }
+
+    /**
+     * Stage-1 LP fallback: a weighted-external-quantity LP opens a starting support, then a
+     * deterministic deletion filter closes gates one by one (sources first, thinnest flow first)
+     * while feasibility holds. The result is a MINIMAL support - no proper subset is feasible -
+     * in a handful of fast LP solves and with no big-M anywhere.
+     */
+    private static Set<Integer> deletionFilter(final Ctx ctx, final double[] floors) {
+        final StageSolve lp0 = solveExternalsLp(ctx, floors, null);
+        if (lp0 == null) return null;
+        Set<Integer> support = lp0.support;
+
+        final double[] gateFlow = new double[ctx.gates.size()];
+        for (int p = 0; p < lp0.externals.length; p++) {
+            gateFlow[ctx.portGate[p]] += lp0.externals[p];
+        }
+        final List<Integer> order = new ArrayList<>(support);
+        order.sort(
+            Comparator.<Integer, Double>comparing(
+                g -> -(ctx.gates.get(g)
+                    .input() ? SRC_WEIGHT : SNK_WEIGHT))
+                .thenComparing(g -> gateFlow[g])
+                .thenComparing(g -> g));
+
+        for (final int gate : order) {
+            if (!support.contains(gate)) continue; // already dropped via a shrunken flow support
+            final Set<Integer> trial = new HashSet<>(support);
+            trial.remove(gate);
+            final StageSolve solved = solveExternalsLp(ctx, floors, trial);
+            if (solved != null) {
+                support = solved.support;
+            }
+        }
+        return support;
+    }
+
+    /**
+     * LP over the gate structure with no binaries: externals outside {@code support} are closed
+     * (null = all open), objective = weighted external quantity (source flow costs fractionally
+     * more than sink flow, mirroring the stage-1 preference).
+     */
+    private static StageSolve solveExternalsLp(final Ctx ctx, final double[] floors, final Set<Integer> support) {
+        final Handles h = ctx.buildModel(DEFAULT_BIG_M, false, floors);
+        for (int p = 0; p < ctx.connectedPorts.size(); p++) {
+            if (support != null && !support.contains(ctx.portGate[p])) {
+                h.extVars[p].upper(0);
+            } else {
+                h.extVars[p].weight(
+                    ctx.gates.get(ctx.portGate[p])
+                        .input() ? SRC_WEIGHT : SNK_WEIGHT);
+            }
+        }
+        final Optimisation.Result result = h.model.minimise();
+        if (!isUsable(result)) return null;
+        return StageSolve.from(ctx, h, result.getValue());
+    }
+
+    /** Stage 1 exact MILP: minimize weighted open-gate count under an upper-bound cut. */
+    private static StageSolve solveStage1(final Ctx ctx, final double[] floors, final List<Set<Integer>> cuts,
+        final Double upperBoundCost) {
+        double bigM = DEFAULT_BIG_M;
+        for (int growth = 0; growth <= MAX_M_GROWTHS; growth++) {
+            final Handles h = ctx.buildModel(bigM, true, floors);
+            h.model.options.time_abort = MILP_CERT_BUDGET_MILLIS;
+            h.model.options.time_suffice = MILP_CERT_BUDGET_MILLIS;
+            final Expression ub = upperBoundCost == null ? null : h.model.addExpression("ub_cut");
+            for (int g = 0; g < ctx.gates.size(); g++) {
+                final double weight = ctx.gates.get(g)
+                    .input() ? SRC_WEIGHT : SNK_WEIGHT;
+                h.gateVars[g].weight(weight);
+                if (ub != null) {
+                    ub.set(h.gateVars[g], weight);
+                }
+            }
+            if (ub != null) {
+                ub.upper(upperBoundCost + 0.5);
+            }
+            // Externals carry an epsilon cost so the solver anchors them at their minimal values:
+            // costless externals can sit at arbitrary vertex values, which both presses the big-M
+            // cap spuriously and pollutes the flow-derived gate support with junk flows. The
+            // epsilon (<= 1e-9 * bigM = 1e-3 per external) can never flip a 1-unit gate decision.
+            for (final Variable ext : h.extVars) {
+                ext.weight(1e-9);
+            }
+            addNoGoodCuts(h, cuts);
+            final Optimisation.Result result = h.model.minimise();
+            if (!isUsable(result)) return null;
+            if (pressesCap(h, bigM)) {
+                bigM *= 10;
+                continue;
+            }
+            // OPTIMAL or DISTINCT (unique optimum) both certify; FEASIBLE = timeout incumbent.
+            return StageSolve.from(
+                ctx,
+                h,
+                result.getValue(),
+                result.getState()
+                    .isOptimal());
+        }
+        return null;
+    }
+
+    /** Stage 2 on the uncertified path: fixed support, plain LP, minimize external quantity. */
+    private static StageSolve solveStage2Fixed(final Ctx ctx, final double[] floors, final Set<Integer> support) {
+        final Handles h = ctx.buildModel(DEFAULT_BIG_M, false, floors);
+        for (int p = 0; p < ctx.connectedPorts.size(); p++) {
+            if (support.contains(ctx.portGate[p])) {
+                h.extVars[p].weight(1.0);
+            } else {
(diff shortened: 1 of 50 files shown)


sbancuz:
Is it faster than doing a simple LP program using ojalgo? I really don't want to maintain a LP solver because I don't think I want to deal with all the edge cases that might come up. Also the examples you provided don't show the 12:1 that you put in the comment so I'm a bit confused

EDIT: also I would like to merge everything into the dev branch so I would prefer that this will be merged onto there so we can have a single branching point

OrderedSet86:
Discussed on Discord:
- It is using ojalgo still, `AutoBalancer.java` is just long because it is full of heuristics. The goal of AUTO mode is to provide a reasonable "just works" default to users. I may switch from ojalgo if I find a MILP solver library that is better at "just works"
- For the 12:1 ratio, 18.46/s fusion heavy naq / 1.54 tiberium heavy naq = 12
- Yep it should merge into balancer, that was the intent
- This is still a WIP PR, there are things I want to fix with it before it's ready to review

OrderedSet86:
requested summary of the AutoBalancer solver (AI gen)
----
## Model

- **Input** = the chart + its pins. Pins, strongest first: explicit extent pins (crafts/s, e.g.
  from a target rate: `extent = targetRate / perCraftQty`), per-node target output rates
  (largest implied extent wins — parallel outputs share one extent), fixed machine counts
  (`extent = count * 20 / durationTicks`). No pin anywhere → refuse with `NO_PIN`: the model is
  homogeneous, every solution scales freely, so any numbers would be an invented anchor.
- **Variables**: one extent per machine (crafts/s), one flow per drawn edge (ingredient/s), one
  external per connected port. Per connected port, one conservation row:
  `sum(edge flows) + external = extent * perCraftQty`, scaled by `1/max(1, qty)` so coefficient
  magnitudes stay near 1 across the 0.05-dust-to-20000-L quantity range.
- **Terminals vs gates**: unconnected ports are free terminals — supplying oil / collecting
  product is the point of the chart, not a cost. Connected ports get *gated* externals. Gate
  granularity is per ingredient per direction, where ingredient identity is structural:
  connected components of ports under the drawn edges (union-find). All input-side externals of
  one component share one SOURCE gate, all output-side one SINK gate — binary count is roughly
  2x the intermediate count, not one per port.
- Edges whose saved port index no longer exists (recipe changed between saves) are dropped, not
  indexed.

## Stage pipeline (one `runStages` pass)

**Fast path.** Stage-3 LP with every gate closed. Feasible → done: 0 gates and 0 external
quantity are trivially optimal, and this LP *is* stage 3.

**Stage 1 — minimize open-gate count.** Sinks weigh 1024, sources 1025, so equal counts prefer
"discard the excess" over "import an intermediate". Two runs:

1. *Deletion filter*: a weighted-external-quantity LP opens a starting support, then gates
   close one by one (sources first, thinnest flow first) while feasibility holds. Result is a
   minimal support — no proper subset feasible — from a handful of LPs, with no big-M anywhere.
2. *Exact MILP slice* (500ms budget): binaries with big-M gate links (M grows x10, up to 3
   times, whenever an external presses the cap), plus an epsilon cost on externals so they
   anchor at minimal values. Certifies or beats the filter; over budget → keep the filter
   answer, marked uncertified in the solution notes.

**Stage 2 — minimize total external quantity**, weighted gate count capped at what stage 1
actually used. The cap covers gates *carrying* flow, however small — capping below that leaves
stage 2 infeasible on a chart stage 1 just solved. Certified path: MILP under the cap.
Uncertified path, or when the free search does not close: plain LP with the stage-1 gates held
fixed, which stage 1's own solution already satisfies.

**Tie enumeration** (certified path only): up to 5 supports tied at (gate count, quantity),
found via no-good cuts.

**Stage 3 — minimize total internal flow**, gates fixed to each candidate support, external
quantity capped at stage 2's optimum with a small relative slack. Pins the free circulation of
fully-recycling loops to a finite value and resolves stage-2 ties deterministically; the least
internal flow wins.

Gate supports are always derived from *flows*, never from binary values — big-M times
integrality tolerance leaks otherwise. The zero-flow threshold is relative to the solution's
own scale, since the model is homogeneous.

## Around the stages

- **Stage 0 — "every machine runs"** — is a second pass, not a starting constraint: pass 1 runs
  floor-free; if machines wired to a pin sit idle, rerun with extent floors set 1000x below
  pass 1's smallest running rate, so a floor can never bind above a rate the chart already
  achieves (a fixed a-priori floor can exceed a machine's natural rate and conjure phantom
  externals). Components containing no pin are exempt — nothing anchors their scale. Failure
  reads "N machines cannot run", never a silent plan with dead machines in it.

src/test/java/com/sbancuz/plannh/RateDisplayTest.java
```diff
@@ -0,0 +1,54 @@
+package com.sbancuz.plannh;
```
sbancuz, line 1: This file is not testing anything

src/main/java/com/sbancuz/plannh/data/flowchart/Balancer.java
```diff
@@ -34,14 +39,62 @@ public String displayName() {
...
+            solution.machineCounts()
+                .size(),
+            solution.openGates(),
+            solution.wallMillis());
+        return buildResultFractional(graph, solution.machineCounts(), solution.notes());
+    }
+
+    /** A quantity-free result: recipes and durations only, plus the reason as a summary note. */
+    @Nonnull
+    private static BalanceResult unbalanced(final Graph graph, final String note) {
```
sbancuz, line 102: There is already balanceNone

src/main/java/com/sbancuz/plannh/data/flowchart/Balancer.java
```diff
@@ -62,30 +115,49 @@ private static BalanceResult balanceNone(final Graph graph) {
...
+    /**
+     * Builds the balance result from (possibly fractional) machine counts. Everything - the
+     * displayed count, the effective rates, the totals - derives from the exact fractional
+     * count, so every displayed number is auditable against every other. A fractional count
+     * reads as that machine's duty cycle; rounding up for physical placement is the reader's
+     * one-step mental operation, not something the math does behind their back.
+     */
+    @Nonnull
+    static BalanceResult buildResultFractional(final Graph graph, final Map<UUID, Double> machineCounts,
+        final List<String> notes) {
```
sbancuz: Why are notes here?

src/main/java/com/sbancuz/plannh/data/flowchart/Balancer.java
```diff
@@ -62,30 +115,49 @@ private static BalanceResult balanceNone(final Graph graph) {
 
     @Nonnull
     static BalanceResult buildResult(final Graph graph, final Map<UUID, Integer> ops) {
+        final Map<UUID, Double> fractional = new HashMap<>(ops.size());
+        for (final Map.Entry<UUID, Integer> entry : ops.entrySet()) {
+            fractional.put(entry.getKey(), (double) entry.getValue());
+        }
+        return buildResultFractional(graph, fractional, List.of());
```
sbancuz: This makes everything work as fractional, which is fine. But it's really conflicting with the integer expectations set as a whole. All these rounding will surely cause some inconsistencies. Also `buildResultFractional` deosn't need to exist, it's the same method as this

src/main/java/com/sbancuz/plannh/gui/FlowchartScreen.java
```diff
@@ -400,11 +402,28 @@ private int computeHeight(final Summary summary, final BalanceResult br) {
             if (br.totalDurationTicks() > 0) {
                 h += SECTION_H;
             }
+            if (!br.notes()
```
sbancuz: Again, what are these notes?

src/main/java/com/sbancuz/plannh/gui/FlowchartScreen.java
```diff
@@ -411,6 +411,32 @@ private int computeHeight(final Summary summary, final BalanceResult br) {
...
         }
 
+        /**
+         * Products/inputs whose amount survives display rounding: netting float residue would
+         * otherwise print as a "0mB/s" line.
+         */
+        private Summary displayedSummary(final Summary s, final BalanceResult br) {
+            final boolean isCycle = summaryMode() == SummaryMode.CYCLES;
+            final float cycleSecs = summaryCycleSecs(br);
+            return new Summary(
```
sbancuz, line 601: It doesn't really matter since it will be rewritten later, but please be mindful of these allocations that don't need to happen.

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
```diff
@@ -66,8 +66,14 @@ public final class AutoBalancer {
...
      * and marked uncertified instead of burning the full stage budget.
+     *
+     * <p>
+     * The number is small because the search is all-or-nothing: across the corpus every proof
+     * that lands does so within 150ms, and the charts that time out still time out at 5s. Paying
+     * more only lengthens the freeze - 5s cost palladium_line 14.0s per solve against 1.2s here,
+     * for identical gate counts, external quantities and machine counts.
      */
-    private static final long MILP_CERT_BUDGET_MILLIS = 5_000;
+    private static final long MILP_CERT_BUDGET_MILLIS = 500;
```
sbancuz: This would be best suited as an "accuracy" setting for all the balancers. (High: 500, Medium: 100, Low: 25) For example, the numbers are just random and arbitrary, but are just to make a point

src/main/java/com/sbancuz/plannh/gui/FlowchartScreen.java
```diff
@@ -186,7 +186,11 @@ public static FlowchartScreen create() {
                             new ButtonWidget<>().overlay(IKey.str("Ops"))
                                 .onMousePressed(_ -> {
                                     final Graph g = canvas.getGraph();
-                                    if (g.getBalanceMode() != BalanceMode.NONE) {
+                                    // AUTO reports exact per-second rates and ignores opsMode, but
+                                    // the summary would still rescale to per-cycle totals, netting
+                                    // recycled ingredients into phantom lines.
+                                    if (g.getBalanceMode() == BalanceMode.OUTPUT
```
sbancuz: This way the changes made by the AUTO won't get saved right?

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
```diff
@@ -478,6 +479,25 @@ private static void addNoGoodCuts(final Handles h, final List<Set<Integer>> cuts
...
+    /**
+     * ojAlgo's branch-and-bound prints every integrality drift it sees ("Obviously infeasible
+     * value ...") straight to the static {@link BasicLogger#ERROR} - roughly 150 lines per solve
+     * on a medium chart. {@code Optimisation.Options.validate} does not gate it: NodeKey passes
+     * that flag as a literal true. Big-M drift is expected here and already handled by deriving
+     * the gate support from flows rather than from the binaries, so the stream is silenced for
+     * the duration of our own solves and restored afterwards - a global assignment would also
+     * swallow the errors of any other mod using ojAlgo, which is not relocated in this jar.
+     */
+    private static Optimisation.Result minimise(final ExpressionsBasedModel model) {
```
sbancuz: The comment above is not really clear, is this because ojAlgo would just print tons of stuff? We can just disable it globally and make a mod setting to enable it for debugging

src/main/java/com/sbancuz/plannh/data/flowchart/Node.java
```diff
@@ -46,6 +47,14 @@ public class Node {
     @Setter
     private boolean machineCountFixed;
 
+    /**
+     * Target production rates by output port index, in ingredient units per second. A target is
+     * a pin: AUTO holds the machine's extent so the targeted output hits the rate exactly, and
+     * the rest of the chart follows. With several targets on one machine the largest implied
+     * extent wins - parallel outputs share one extent, so only the tightest can be exact.
+     */
+    public final Map<Integer, Double> targetOutputRates = new LinkedHashMap<>();
```
sbancuz, line 56: This concept is cool, but it's either this or the machineCountFixed. I'm fine with having this approach, but I don't want 2 ways of doing things

src/main/java/com/sbancuz/plannh/gui/RecipeNodeWidget.java
```diff
@@ -511,11 +511,18 @@ private String portLabel(final Port<?> port, final int index, final NodeBalance
...
             // operation count showed rounded rates next to exact input rates on the same node.
             final float total = effectiveTotal(nb, index, output, fs.amount);
-            return GuiHelper.formatRate(total / sec) + "/s " + fs.getLocalizedName();
+            return GuiHelper.formatRate(total / sec) + "/s " + portDisplayName(port);
         }
         return null;
     }
 
+    @Nullable
+    private static String portDisplayName(final Port<?> port) {
```
sbancuz: The properties have this formatter embedded directly (they should if they don't, maybe it's in another branch idk)

src/main/java/com/sbancuz/plannh/data/flowchart/Serializer.java
```diff
@@ -205,7 +205,12 @@ private static JsonObject graphToJson(final Graph graph) {
             obj.addProperty("x", node.x);
             obj.addProperty("y", node.y);
             obj.addProperty("machine", node.machineName);
-            obj.add("recipeId", node.recipeId.toJsonObject());
+            // Null-tolerant on both sides: a node whose recipe never resolved must not make the
+            // whole slot unsaveable, and everything downstream already treats a null recipeId as
+            // "handler unavailable" (RecipeHandlerRef.of is null-safe).
+            if (node.recipeId != null) {
```
sbancuz, line 279: Can this happen? At this point why even saving this node

src/main/java/com/sbancuz/plannh/gui/GuiHelper.java
```diff
@@ -40,7 +40,8 @@ public static String trimTrailingZeros(final String s) {
      * trickle rates that fewer digits would round to an all-zero line.
      */
     public static String formatRate(final float rate) {
-        if (rate >= 1000000000f) return String.format("%.1fB", rate / 1000000000f);
+        // G, not B: fluid amounts already use B for buckets, and that is how players read it.
+        if (rate >= 1000000000f) return String.format("%.1fG", rate / 1000000000f);
```
sbancuz: I want to keep it consistent with the rest of the pack, haven't opened a gtnh instance in quite a while, is this how is displayed in AE2? And GT for that matter

src/main/java/com/sbancuz/plannh/data/provider/gregtech/GTProvider.java
```diff
@@ -282,31 +282,31 @@ public Map<RecipeProperty<?>, Object> extract(final @NotNull Node node, final @N
                 new Port<>(
                     RecipePropertyAPI.ITEM,
                     r.mInputs[i],
-                    r.mInputChances != null ? r.mInputChances[i] / 100.0f : 1.f));
+                    r.mInputChances != null ? r.mInputChances[i] / 10000.0f : 1.f));
```
sbancuz: Can you make it a static constant with a clear comment, I already know I will forget this because I already changed it from 10000 to 100 because I thought it was stupid

src/main/java/com/sbancuz/plannh/data/provider/gregtech/GTProvider.java
```diff
@@ -275,37 +275,39 @@ public Map<RecipeProperty<?>, Object> extract(final @NotNull Node node, final @N
         node.inputs.clear();
         node.outputs.clear();
 
-        // GT recipe arrays may contain null slots (e.g. gap outputs); skip them.
+        // GT recipe arrays may contain null slots (e.g. gap outputs); skip them. Stacks are
+        // copied because Port.merge mutates amounts in place - wrapping the live recipe stacks
+        // corrupts the shared GT recipe on every extraction.
         for (int i = 0; i < r.mInputs.length; i++) {
```
sbancuz, line 284: Revert e0a9fdcf700e8c4a49690df85db09567c3204ad7, for the GT fixes don't need to copy itemstacks and the chance formatting is wrong

sbancuz (changes requested):
I haven't looked at AutoBalancer yet since it's the most complicated part. These are all comments around that. Will have a new review just for the balancer in a bit

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
```diff
@@ -0,0 +1,1160 @@
...
+        final List<ConnectedPort> connectedPorts = new ArrayList<>();
+        final Map<Long, Integer> portLookup = new HashMap<>(); // (machine, port, input) -> index
+        final List<Gate> gates = new ArrayList<>();
+        int[] portGate; // connected port index -> gate index
+        int[] portComponent; // connected port index -> ingredient component root
+        boolean anyPin;
+        final List<String> notes = new ArrayList<>();
+
+        Ctx(final Graph graph, final Map<UUID, Double> extraExtentPins) {
+            final List<Node> nodes = new ArrayList<>(graph.getNodes());
```
sbancuz: This is not the first place where we need a particular ordering of values from an hashmap. I think a better (and easier) approach would be to migrate to a `SortedMap` (even a custom one that extends one from fastUtils) 

This comments is for all the things in the graph since even edges receive this treatment

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
```diff
@@ -0,0 +1,1160 @@
...
+    /** Suffix of every wiring-diagnostic note; tests key on it. */
+    public static final String MISSING_EDGE = "missing an edge?";
+
+    /** A port on a specific machine. {@code input} distinguishes the two port lists. */
+    public record PortRef(UUID nodeId, int portIndex, boolean input) {}
+
+    /** An external attached to a port, in ingredient units per second. */
+    public record External(PortRef port, double ratePerSecond) {}
+
+    public record Solution(Map<UUID, Double> extentsPerSecond, Map<UUID, Double> machineCounts,
```
sbancuz, line 322: Basically everything here is just used for testing, these are a lot of allocations that don't amount to anything in game

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
```diff
@@ -0,0 +1,1160 @@
...
+        Map<UUID, Double> edgeFlowsPerSecond, List<External> gatedSources, List<External> gatedSinks,
+        List<External> terminalInputs, List<External> terminalOutputs, int openGates, double externalQuantity,
+        double totalInternalFlow, boolean floorsUsed, long wallMillis, List<String> notes) {}
+
+    public record Result(Solution solution, String failure) {
```
sbancuz, line 327: This is not rust, while I like this type of pattern maybe consider using exceptions for one-off error messages or pass a list of strings to populate if you need to collect a lot of errors

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
```diff
@@ -0,0 +1,1160 @@
...
+    // ---------------------------------------------------------------------------------------
+
+    private record ConnectedPort(int machine, int portIndex, boolean input, double qtyPerCraft, List<Integer> edges) {}
+
+    private record EdgeData(UUID id, int srcPort, int dstPort) {}
+
+    /** One gate: an ingredient component in one direction, covering the listed ports. */
+    private record Gate(boolean input, List<Integer> ports) {}
+
+    private static final class MachineData {
```
sbancuz: This is just a copy for no reason of a node

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
```diff
@@ -0,0 +1,1160 @@
...
+        return false;
+    }
+
+    // ---------------------------------------------------------------------------------------
+    // Model context
+    // ---------------------------------------------------------------------------------------
+
+    private record ConnectedPort(int machine, int portIndex, boolean input, double qtyPerCraft, List<Integer> edges) {}
+
+    private record EdgeData(UUID id, int srcPort, int dstPort) {}
```
sbancuz: This is just an edge

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
```diff
@@ -0,0 +1,1160 @@
...
+            if (v != null && v.doubleValue() > 0.9 * bigM) return true;
+        }
+        return false;
+    }
+
+    // ---------------------------------------------------------------------------------------
+    // Model context
+    // ---------------------------------------------------------------------------------------
+
+    private record ConnectedPort(int machine, int portIndex, boolean input, double qtyPerCraft, List<Integer> edges) {}
```
sbancuz: If I understood correctly a port is just a compact representation of an edge, just store int edges[2]

sbancuz (changes requested):
I think I understand how this model works:
- fastpath -> try stage 3 (the full LP one)
- if it doesn't find a solution, so you need external stuff
    - maximise the usage of internal flow
    - minimise the external reliance
    - minimize flow 

While I think this idea can work out well, while I'm still not fully convinced all this complication is needed since I don't really like implicit stuff (I do understand sinks/sources are not in this branch). If a user doesn't specify that the base can provide 300 wood/s to supplement the line, to me it's their fault. I hope in the future I change my mind.

That aside the code for the balancer has tons of duplicated state and model (like what an edge is) and it's re-created on every balancing run. Since in this branch this behavior is still performed constantly, it's not ideal (I think I will support rebalancing on button press as well as on every graph change). Also the code leaves a bit to be desired, but I do understand it's LLM code...

src/main/java/com/sbancuz/plannh/data/flowchart/Node.java
OrderedSet86: Sometimes you can only afford 1 expensive multiblock, and sometimes you arbitrarily want to decide on making 1000L/s of something. Both are valid and I used both all the time before (as did others)

src/main/java/com/sbancuz/plannh/data/flowchart/Serializer.java
OrderedSet86: Yes, this was added to handle a real crash I encountered. It may be dev-only crash handling rather than end user handling though

src/main/java/com/sbancuz/plannh/data/provider/gregtech/GTProvider.java
OrderedSet86: 1. If by chance formatting you mean `" (" + Math.round(port.getAmount() * port.getChance() * 100) + "%)"`, the chance formatting is built to handle this palladium salt dust sifter, which has 9 outputs of the same item, with 8 at 10% and 1 at 15%. The chance formatting correctly inputs 95% on my branch, before it would output 10%
Image

2. The GT itemstack copies were because for whatever reason, the sifter chance calculation was stale on a slot even after I removed/added the node and after I restarted the game. After adding the copy the stale state problem went away. This may become relevant when we make correctness fixes in the future - we don't want people to have to remake their graph after a patch. Not sure what the cause is, some kind of aggressive slot scoped cache?

src/main/java/com/sbancuz/plannh/gui/FlowchartScreen.java
OrderedSet86: They're debug information output by the solver, e.g.

Image

src/main/java/com/sbancuz/plannh/gui/GuiHelper.java
OrderedSet86: MUI2 has its own way of displaying these values (as indicated in the comment) but it has some funky rounding rules I didn't like (0.0001 -> display as 0)

src/test/java/com/sbancuz/plannh/RateDisplayTest.java
OrderedSet86: This is testing the behavior of the rounding formatter, but I agree that this test isn't that useful

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
OrderedSet86: This is a dead comment, ojAlgo on 47.3.1 used to print spam and throw a bunch of instability errors, but it's fine on 57.1.0

src/main/java/com/sbancuz/plannh/data/flowchart/Node.java
sbancuz: The UI can show and represent whatever it wants, but in the backend i'd like to have one way to do something. 1 machine can be represented by setting targets automatically. The point still stands having both the "fixed" that only works with the first 2 modes and target that only works with the AUTO is confusing even from the perspective of the user

src/main/java/com/sbancuz/plannh/data/flowchart/Serializer.java
sbancuz: Then sure, skip directly serialization in this case. We don't care about recipes that we can't ever lookup

src/main/java/com/sbancuz/plannh/gui/GuiHelper.java
sbancuz: Sorry, maybe I wasn't clear enough, I was referring to B -> G for indicating billions

src/main/java/com/sbancuz/plannh/data/provider/gregtech/GTProvider.java
sbancuz: It may seem confusing at first glance, but this should handle everything 
```java
// Port.java
    public void merge(final Port<?> other) {
        final int newAmount = getAmount() + other.getAmount();
        this.chance = (this.getAmount() * this.chance + other.getAmount() * other.chance) / newAmount;
        type.setAmount(value, newAmount);
    }
```

Also I've not ran the math, but surely it's not 95% to yield something in that recipe

OrderedSet86: image

Discussed on Discord, sifters can multi output so the naive 10% * 8 + 15% = 95% is correct

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
OrderedSet86: Done

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
OrderedSet86: I implemented this as a `solverEffortPercent` in Config, it is implemented as a multiplier on a bunch of these MILLIS values (default 100)

src/main/java/com/sbancuz/plannh/data/flowchart/Balancer.java
OrderedSet86: LLM response:

> These two aren't the same result, though I agree there were two builders where one would do.
>
> `balanceNone` emits the user's stored machine counts; `unbalanced` emits a quantity-free result —
> zero operations, zero total duration, no property totals — which is what the summary panel reads
> as "this chart is wiring, there are no numbers yet". I tried routing `unbalanced` through the
> shared builder with an all-zero count map and it changes what's on screen:
> `buildResultFractional` sets `totalDurationTicks` to `max(durPerOp)` instead of 0 and fills
> `propertyTotals` with zero entries, both of which the panel renders.
>
> What I did fix is the duplication that *was* real — see `buildResultFractional` thread.

I think the contention is differences between design in AUTO and INP/OUT. I would prefer if you handled this because it depends on what you want to do with INP/OUT after this PR

src/main/java/com/sbancuz/plannh/data/flowchart/Balancer.java
OrderedSet86: LLM response:

> Agreed on the duplication, though it was the other one: `buildResult` was a pure
> `Map<UUID,Integer>` → `Map<UUID,Double>` adapter with no logic of its own, and its signature had
> drifted to 2 params against `buildResultFractional`'s 5. It's deleted; the two callers build the
> map directly and there's one builder now.
>
> On fractional-vs-integer, I'd like to keep the split. AUTO's answer genuinely is fractional —
> 0.533 of an LCR is a real ratio, and rounding before display makes the ratios stop matching each
> other. The ILP modes round up at the point they extract counts because they're placing real
> machines. All the rounding you're worried about is in one place, `balanceILP`'s
> `Math.ceil(raw - 1e-6)`; everything downstream of it is the exact number.

Like the last comment, I think this comes down to your future design intent for INP/OUT. I would rather not change behavior, since those are your solvers

src/main/java/com/sbancuz/plannh/gui/FlowchartScreen.java
OrderedSet86: Resolved, LLM response:

> Fixed, and it was worse than it looked: `wrapNotes` runs the font renderer over every note and was
> called **twice per frame** — once from `computeHeight` to measure the panel, again from `draw`.
> It's memoized against the `BalanceResult` that produced it now, using the same "has this been
> re-solved" identity check the arrow router already uses.
>
> The bigger version of this is also gone: the solve itself is memoized on `Graph` behind a dirty
> flag (`balance`/`summary`/`boundary`/`choices`), so a steady-state frame does no solver work at
> all. It used to re-solve whenever anything asked for the boundary chips.

src/main/java/com/sbancuz/plannh/data/flowchart/Node.java
OrderedSet86: AUTO supports fixed machine counts now. If you try setting both fixed machine and target on a node, it picks target as a tie breaker.

For the residual part of this problem, this is another INP/OUT legacy handling problem. INP/OUT doesn't understand what a target is, and I don't want to make changes to it as I don't understand those solvers' design intent very well.

src/main/java/com/sbancuz/plannh/data/provider/gregtech/GTProvider.java
OrderedSet86: Done:
```java
/** GT5u stores a chance as 1..10000, where 10000 is 100%. */
private static final float GT_CHANCE_SCALE = 10_000f;
```

src/test/java/com/sbancuz/plannh/RateDisplayTest.java
OrderedSet86: Actually I've come around on this test, correct behavior of the rounding is important as it touches userspace. I had to iterate on this behavior based on user feedback a few times in gtnh-flow, so getting it right is important

src/main/java/com/sbancuz/plannh/gui/FlowchartScreen.java
OrderedSet86: LLM response:

> Right, and I think the thing worth pinning down is that **`machineCount` means something different
> depending on the mode** — that's the part that's genuinely confusing, and it's not documented
> anywhere a reader would find it.
>
> - **AUTO: `machineCount` is input only.** The solve's answer is fractional (0.533 of an LCR) and
>   lives only in the `BalanceResult`, which is derived state recomputed whenever the chart goes
>   dirty. Nothing about the answer is written to the graph or saved. What *is* saved is the chart
>   plus its pins — `machineCount`, `machineCountFixed`, `targetOutputRates` — the mode, and the
>   user's tie-break (`excessChoice`, through `recordEdit`, so undo covers it).
> - **INPUT/OUTPUT: `machineCount` is input *and* output.** `balanceILP` ceil-rounds its answer and
>   writes it straight back into `node.machineConfig` — deliberately, per the comment there, "so
>   they're available when balancing is later switched off".
>
> So switching mode changes whether solving edits your chart. AUTO can't write back even if we
> wanted it to: `machineCount` is an `int` and the answer isn't, so writing it would round away the
> ratios that are the whole point, and the rounded value would then be read back as the input to the
> next solve.
>
> **That asymmetry has a real bug in it, on the ILP side.** `balanceILP`'s writeback runs inside
> `graph.balance()`, which is called from `SummaryWidget.draw()` — so in OUTPUT/INPUT mode, merely
> *looking* at a chart rewrites its machine counts. It isn't wrapped in `recordEdit`, so undo can't
> reach it, and the next unrelated `PlanAPI.save()` persists it. I haven't touched it because it's
> orthogonal to this PR, but if you want it fixed here say so — the options are roughly: move the
> writeback out of the solve to an explicit user action, or keep it and bracket it properly.
>
> The one thing I did change nearby: the Ops button was firing for every mode, so AUTO was saving a
> flag it ignores. It's guarded by `BalanceMode.usesOpsMode()` now.

src/main/java/com/sbancuz/plannh/gui/GuiHelper.java
OrderedSet86: | mod | convention |
| --- | --- |
| NEI `codechicken.nei.util.ReadableNumberConverter` | kMGTPE, base 1000 |
| AE2 `appeng.util.ReadableNumberConverter` | kMGTPE, base 1000 — identical class |
| GT `GTUtility.formatNumbers` | `DecimalFormat` grouping, 1,234,567, no suffixes |

I have modified the number display to follow NEI/AE2 convention globally

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
OrderedSet86: These cost microseconds (0.001% of the solver time) and greatly increase the ability to test and extend the solver in the future. I wouldn't recommend touching them. In particular, these ended up getting used for the source/sink chip nodes:

> `gatedSources`/`gatedSinks`/`terminalInputs`/`terminalOutputs` now drive the boundary chips on the canvas (`BalanceView.boundary()` → `CanvasWidget.drawExternalChips()`), so those four are in-game data now, not test scaffolding. `machineCounts` backs every displayed number.

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
OrderedSet86: LLM response:

> Taking the second half first, because I think it's already the design: there are two channels, and they split exactly along the line you drew.

> A list of strings, for everything non-fatal. `Solution.notes` is assembled from three sources (`FlowModel.toSolution`) — the attempt's own notes, `wiringDiagnostics` which returns a `List<String>` with one entry per suspicious ingredient, and the overshot-target notes. Five producers across the solver, all appending. That is the "pass a list of strings to populate" mechanism, and it's what carries "'X' imports Y externally but the chart also produces it at 'Z' — missing an edge?" to the summary panel.

> One string, for the fatal case — and multiple fatal errors can't coexist. All 9 failure sites are early returns down a sequential pipeline: stage 2 cannot fail if stage 1 produced no support, and nothing continues past a failure to accumulate a second one. Where a single failure has several causes, they're already collected into a list and joined — `diagnosePins` drops pins weakest-first and reports `"these pins cannot hold together …: 'A' (target rate), 'B' (fixed machine count)"`. And where a nested failure needs context, the outer frame adds it: `first.failure + " - " + conflict`, or `idleCount + " machines cannot run: " + floored.failure`.

> So the shape isn't "Rust errors instead of exceptions" so much as "one fatal reason, many non-fatal notes", which I'd argue is right for a solver whose normal output includes advice.

> On exceptions specifically, the thing that makes me want to keep the value at the public boundary is `NO_PIN`: an unpinned chart isn't an error, it's the normal state of a chart you're still drawing, and the one consumer branches on exactly that —

```java
if (AutoBalancer.NO_PIN.equals(result.failure())) return unbalanced(graph, result.failure());
```

> Making that an exception means throwing on the common path and needing two exception types to re-encode a distinction the record states plainly.

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
OrderedSet86: 1. A port is a fan out point for an arbitrary number of edges
2. This also carries `qtyPerCraft`, the per-craft quantity after `MachineConfig` multipliers and chance. That's the actual LP coefficient and it can't live on `Port`, because it depends on node-level config rather than on the port.

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
OrderedSet86: LLM response:

> Accurate as a description — it is `Edge` with the two endpoints resolved to interned port indices,
> keeping `id` only so the solution can key edge flows by the graph's own UUID. I went to replace it
> with parallel arrays in the style of the `int[] portGate` / `int[] portComponent` next to it, and
> it comes out worse. Writing both out:
>
> ```java
> // now
> final List<EdgeData> edges = new ArrayList<>();
> record EdgeData(UUID id, int srcPort, int dstPort) {}
>
> final int e = edges.size();
> edges.add(new EdgeData(edge.id, srcPort, dstPort));
> for (final EdgeData e : edges) union(root, e.srcPort(), e.dstPort());
> for (int e = 0; e < edges.size(); e++) flows.put(edges.get(e).id(), attempt.flows[e]);
> ```
> ```java
> // as arrays
> int[] edgeSrcPort;
> int[] edgeDstPort;
> UUID[] edgeId;
> int edgeCount;          // because length != count
>
> edgeSrcPort[edgeCount] = srcPort;
> edgeDstPort[edgeCount] = dstPort;
> edgeId[edgeCount] = edge.id;
> final int e = edgeCount++;
> for (int e = 0; e < edgeCount; e++) union(root, edgeSrcPort[e], edgeDstPort[e]);
> for (int e = 0; e < edgeCount; e++) flows.put(edgeId[e], attempt.flows[e]);
> ```
>
> That `edgeCount` is the problem. The build loop skips dangling edges — ones whose saved port index
> no longer exists because a recipe lost an output — so the count isn't known until the loop ends.
> `portGate` and `portComponent` get to be `new int[n]` precisely because they are allocated after
> `connectedPorts` is complete; this one can't. So the array form has to over-allocate to
> `graph.getEdges().size()` and carry a count that has to be used instead of `.length` at every site,
> which is three arrays to keep in step plus a length that lies, in exchange for deleting a one-line
> record.
>
> The allocation it would save is also the same order as the edge-flow map in the other thread: 604
> records on nanocircuits, against an 800ms solve.
>
> So I'd keep it, unless the objection is specifically that the *name* reads like a second `Edge` —
> `InternedEdge` or `EdgeRow` would say "this is the solver's index-space copy" more honestly, and
> that I'm happy to change.

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
OrderedSet86: LLM response:

> It holds the `Node` by reference — nothing is copied. What it adds splits in two, and only one
> half is really arguable:
>
> - `durTicks`, `inQty[]`, `outQty[]` — the flattened LP coefficients, post-`MachineConfig`
>   multipliers, chance and throughput factor. You're right that these are derivable from the node;
>   they could be cached on `Node` and invalidated on config change. That's a real option.
> - `pinnedExtent` / `pinKind` — these belong to the solve rather than to the chart, for two
>   concrete reasons rather than on principle. First, pins are a **per-call argument**:
>   `AutoBalancer.solve(graph, extraExtentPins)` takes them per invocation, and solving one graph
>   twice with different pins gives different answers from the same unmodified chart — which is how
>   the scale-invariance test drives the same chart at five different pin scales. Second,
>   `diagnosePins` **temporarily nulls and restores them mid-solve** while it works out which pins
>   conflict; doing that to a field on the user's node is a different proposition to doing it to
>   solver scratch.
>
> Checked while writing this: after an AUTO solve the node is untouched — `machineCount` and
> `machineCountFixed` come back exactly as they went in, which is the contract from the other
> thread.
>
> Note this moved out of `AutoBalancer` into `FlowModel` in the split, so the line this thread
> points at doesn't exist any more.

OrderedSet86:
I'm going to bed since it's 1am here, so the main PR body isn't updated, but all comments should be addressed (either for your comments / choices) and the code in a reviewable state

src/main/java/com/sbancuz/plannh/data/flowchart/AutoBalancer.java
sbancuz: Yeah, it may be a simple microptimization, but I'm talking as the owner of a potato pc, every allocation counts. I suppose if we really need all of this information we can make a custom profiler that is just a noop when running the game

PR #28, opened 2026-08-01
Steps

sbancuz:
A step pins an external flow into or out of the network. Its amount is always expressed per second and is treated as a fixed constant.
- Positive amount = source: the step injects N items/fluid per second into the network through its output port.
- Negative amount = sink: the step consumes |N| items/fluid per second through its input port.
- Zero = no effect. The step is ignored even when connected since it's the same as connecting 2 nodes together

In balancing:
- Sinks force the machines feeding them to scale their operator count up until their per-second production covers the sink's demand.
- Sources feed downstream machines, reducing (or fully eliminating, if supply is enough) what those machines demand from upstream producers.
- Step values themselves never change.

In the summary:
- A positive step appears as a product (surplus).
- A negative step nets against products; whatever isn't covered shows up as an External Input (deficit).
- Step-sourced lines display their raw per-second value in both CYCLES and THROUGHPUT modes.


src/main/java/com/sbancuz/plannh/data/flowchart/ItemStackAdapter.java
TheYoingLad, line 1: Isn't there an existing implementation if this somewhere?

TheYoingLad (commented):
Cant comment on the backend stuff.
If the frontend is placeholder then ill approve it, ill take a proper look and possibly rework it later.

src/main/java/com/sbancuz/plannh/data/flowchart/ItemStackAdapter.java
sbancuz: If there is, I can't find it. Look at everything in the project that implements JsonSerializer

sbancuz:
> Cant comment on the backend stuff. If the frontend is placeholder then ill approve it, ill take a proper look and possibly rework it later.

I'd say half and half, could be made prettier, but I suppose that's it no? Do you have another design in mind?

src/main/java/com/sbancuz/plannh/gui/step/AmountTextFieldWidget.java
```diff
@@ -0,0 +1,74 @@
...
+        super.onFocus(context);
+        setTextAlignment(Alignment.CenterLeft);
+
+        setTextColor(PlannhColors.TEXT_BLACK.getColor());
+    }
+
+    @Override
+    protected void onTextChanged() {
+        super.onTextChanged();
+    }
```
TheYoingLad, line 51: Unnecessary override

src/main/java/com/sbancuz/plannh/gui/step/AmountTextFieldWidget.java
```diff
@@ -0,0 +1,74 @@
...
+    public AmountTextFieldWidget(StepWidget parent) {
+        super(
+            parent,
+            parent.getData()
+                .getAmount());
+
+        data = parent.getData();
+
+        width(50);
+        height(16);
```
TheYoingLad, line 24:
```suggestion
        size(50, 16);
```

src/main/java/com/sbancuz/plannh/gui/step/StepAreaWidget.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private static final int SLOT_SIZE = 18;
+
+    private final StepWidget parent;
+    private final Step data;
+
+    public StepAreaWidget(StepWidget parent) {
+        this.parent = parent;
+        this.data = parent.getData();
+        size(60, 60);
```
TheYoingLad, line 29:
```suggestion
        size(60);
```

src/main/java/com/sbancuz/plannh/gui/step/StepAreaWidget.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public boolean handleDragAndDrop(final ItemStack draggedStack, final int button) {
+        data.setFilter(draggedStack);
+        return true;
+    }
+
+    @Override
+    public void draw(final ModularGuiContext context, final WidgetThemeEntry<?> widgetTheme) {
```
TheYoingLad, line 39: When the port widget gets merged, this should be rewritten to use it. That way, this override should hopefully be removed.

src/main/java/com/sbancuz/plannh/gui/step/StepWidget.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public int getInputPortAt(final int localMx, final int localMy) {
+        final int py = getWorldHeight() / 2 - PORT_HALF;
+        if (localMx >= 0 && localMx < PORT_SIZE && localMy >= py && localMy < py + PORT_SIZE) return 0;
+        return -1;
+    }
+
+    @Override
+    public void draw(final ModularGuiContext context, final WidgetThemeEntry<?> widgetTheme) {
```
TheYoingLad, line 105: Same as my comment on the StepAreaWidget.draw() method

TheYoingLad (changes requested):
Only a few nitpicks, happy with everything else.

PR #36, opened 2026-09-04
Add machine sharing groups

Eldrinn-Elantey:
A chain often runs several recipes on the same machine, and the chart already knows the answer: every node's count is machine time, so a recipe filling a third of a machine reads 0.33. What was missing is saying which recipes share the hardware, and being able to plan against it.

A machine group is a group that stands for one machine. It takes in nodes of a single recipe handler, its members share the settings that decide duration, and their counts add up in its header. Giving it a machine count enters the solve as one row per group, the summed machine time held at or below the machines it has, so a chain that does not fit the hardware is reported as such instead of being measured after the fact. Ordinary groups are untouched, and a machine group without a count constrains nothing.

Not covered: minimizing the number of physical machines, which would add a machine-count term to the AUTO objective rather than a constraint.

### DIFF
--- a/src/main/java/com/sbancuz/plannh/gui/GroupWidget.java
+++ b/src/main/java/com/sbancuz/plannh/gui/GroupWidget.java
@@ -2,7 +2,9 @@ package com.sbancuz.plannh.gui;
 
 import static com.sbancuz.plannh.data.flowchart.Group.GROUP_MIN_W;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
@@ -11,15 +13,22 @@ import com.cleanroommc.modularui.api.drawable.IKey;
 import com.cleanroommc.modularui.drawable.DynamicDrawable;
 import com.cleanroommc.modularui.drawable.GuiTextures;
 import com.cleanroommc.modularui.drawable.Rectangle;
+import com.cleanroommc.modularui.screen.RichTooltip;
 import com.cleanroommc.modularui.screen.viewport.ModularGuiContext;
+import com.cleanroommc.modularui.utils.Alignment;
 import com.cleanroommc.modularui.utils.Color;
 import com.cleanroommc.modularui.value.BoolValue;
 import com.cleanroommc.modularui.widgets.ButtonWidget;
 import com.cleanroommc.modularui.widgets.ColorPickerDialog;
 import com.cleanroommc.modularui.widgets.CycleButtonWidget;
+import com.cleanroommc.modularui.widgets.TextWidget;
 import com.cleanroommc.modularui.widgets.ToggleButton;
 import com.cleanroommc.modularui.widgets.layout.Flow;
+import com.sbancuz.plannh.data.flowchart.Graph;
 import com.sbancuz.plannh.data.flowchart.Group;
+import com.sbancuz.plannh.data.flowchart.Node;
+import com.sbancuz.plannh.data.flowchart.balancer.BalanceResult;
+import com.sbancuz.plannh.data.flowchart.balancer.Balancer;
 
 import lombok.Getter;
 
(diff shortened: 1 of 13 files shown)


src/main/java/com/sbancuz/plannh/gui/GroupWidget.java
```diff
@@ -103,6 +119,64 @@ public GroupWidget(CanvasWidget canvas, Group data) {
...
+     * machine sharing or holds no solved nodes, which collapses the widget away.
+     */
+    private TextWidget<?> loadLabel() {
+        return new TextWidget<>(IKey.dynamic(() -> {
+            final List<Map.Entry<String, Double>> load = machineLoad();
+            if (load.isEmpty()) return "";
+            if (load.size() == 1) return "×" + GuiHelper.formatCount(
+                load.get(0)
+                    .getValue());
+            return load.size() + " types";
```
sbancuz: This should be localized

src/main/java/com/sbancuz/plannh/gui/GroupWidget.java
```diff
@@ -67,6 +76,12 @@ public GroupWidget(CanvasWidget canvas, Group data) {
             .reverseLayout();
 
         buttonRow.child(new CloseButtonWidget(this))
+            .child(
+                new ToggleButton().value(new BoolValue.Dynamic(data::isMachineSharing, data::setMachineSharing))
+                    .overlay(
+                        IKey.str("MS")
+                            .color(Color.WHITE.main))
+                    .addTooltipLine("Toggle Machine Sharing: sum the machine counts of the recipes in this group"))
```
sbancuz: ditto

sbancuz (changes requested):
This seems to be just a visual change, from the discord discussion I figured you wanted a more logical change that tied to the balancer in some way

Eldrinn-Elantey:
The machine-time sum should group by machine settings, not just machine name, otherwise two LCR nodes at different tiers get added together. Worth syncing the settings inside a sharing group instead?

sbancuz:
> The machine-time sum should group by machine settings, not just machine name, otherwise two LCR nodes at different tiers get added together. Worth syncing the settings inside a sharing group instead?

I'd say a group can only have one type of machine (don't use the name since they can be renamed, use the recipe handler) and in that group all the settings are synced between all other machines. 

So these special groups will only ever represent one machine avoiding confusion, also could you try to make it a bit more apparent that it's using these special groups? Just to avoid confusion

Eldrinn-Elantey:
Reworked it into a separate machine group type: one handler per group (taken from the handler, not the display name), settings shared between its nodes, and a badge plus a thicker border so it reads differently from a plain group. A node of another handler simply is not taken into the group, silently for now, tell me if you want that visible.

sbancuz:
Sadly we don't have (yet) any type of message popup. but instead of failing silently, look into groupablewidget, there is a function that disallows widgets to be on top of one another, maybe you could change it a bit to make it snap back nodes that shouldn't go into that group.

Eldrinn-Elantey:
Done, a node of another handler now snaps back to where the drag started instead of being left inside the group.

sbancuz:
The last commit I think it would be cleaner if it would've used this:

```java
  // GroupableWidget.java

  @Override
    public boolean canDropHere(int x, int y, @Nullable IWidget widget) {
        return canvas.isMouseInsideCanvas() && canvas.getFlowchartWidgets()
            .values()
            .stream()
            .filter(f -> !(f instanceof GroupWidget) && !dragStartIntersect.contains(f)) <-- This
            .map(AbstractWidget::getArea)
            .noneMatch(area -> area.intersects(getArea()));
    }

    public void removeFromGraph() {
        canvas.getFlowchartWidgets()
            .remove(data.getId());
        dataContainer.remove(data.getId());
    }

    private void adjustGroupMembership() {
        ParentWidget<?> oldParent = (ParentWidget<?>) getParent();
        ParentWidget<?> newParent = StreamSupport.stream(
            getContext().getAllBelowMouse()
                .spliterator(),
            false)
            .filter(w -> (w instanceof FlowchartWidget<?, ?> || w instanceof CanvasWidget) && w != this) <-- This
            .map(w -> (ParentWidget<?>) w)
            .findFirst() // this should always find at least 1 match (the canvas)
            .orElseThrow();
```

Right now it seems that you are re-creating some functionality

Eldrinn-Elantey:
canDropHere and adjustGroupMembership are on GroupableWidget, and RecipeNodeWidget is a plain Widget that drags itself in onMouseDrag, so a node never reaches that path and there was nothing to reuse. Want me to move node dragging onto IFlowchartDraggable so the refusal becomes a canDropHere?

sbancuz:
Ye ye, I got confused between the branches. On the mui2-rework they are...

So it's either this or drop the last commit and leave a to-do

Sorry to have wasted your time

Eldrinn-Elantey:
No time wasted. I would keep it then: without it a refused node just stays in the group silently, and once mui2-rework lands the check moves into canDropHere.

sbancuz (approved):

